PR #168, opened 2026-03-22
fix: add connection timeout and max buffer size limit

Yuan-ManX:


dutifulbob:
## Triage result

### Quick read
- Human decision needed: decide whether to require a concrete reproduction/regression test for this bug fix or accept the remaining validation gap and make the landing decision
- Validation status: `fix_not_proven`
- Initial merge state: conflict detected in `src/queue-ipc.ts` against `origin/main`

### Intent
This PR aims to make queue-owner communication fail fast instead of hanging or growing without bounds by adding a socket connection timeout and a maximum response buffer size.

### Why
The queue IPC path can currently wait forever for a socket connection and can accumulate unbounded incoming data until a newline arrives, which can leave commands stuck and waste memory if the queue owner is stalled or misbehaving.

### Codex review
The change looks right-shaped for the shared queue IPC client path:
- it adds a 5s timeout to queue-owner socket connection attempts
- it adds a 10 MB cap to the common response buffer in `runQueueOwnerRequest`

No Codex review result is available in this run state.
The current blocker is validation: no targeted repro steps or targeted test command could be derived, so the fix is not proven yet.

### CI/CD
No CI/CD result is available in this run state.
No final conflict-gate result is available in this run state.
An initial merge conflict was detected in `src/queue-ipc.ts` against `origin/main`.

### Recommendation
Escalate to a human. The next call is whether to require a concrete reproduction/regression test for the hanging or unbounded-buffer behavior before landing, or to accept the remaining validation gap and make the landing decision manually.

PR #174, opened 2026-03-23
fix: handle -32601/-32602 in session/load fallback

### LINKED ISSUE #152: shouldFallbackToNewSession does not handle -32602 from session/load, causing crash with cursor-agent
## Problem

When using `acpx` with `cursor-agent` (Cursor IDE CLI) in `prompt` mode (persistent sessions), session reconnection fails with:

```
AcpRuntimeError: acpx exited with code 1: code=ACP_TURN_FAILED
```

The `exec` mode works fine because it always creates a new session via `session/new`.

## Root Cause

1. `cursor-agent` reports `agentCapabilities.loadSession = true` during ACP `initialize`
2. When `acpx` tries to restore a previous session via `session/load`, cursor-agent returns JSON-RPC error `-32602` (Invalid params) — its `session/load` implementation appears incomplete
3. `shouldFallbackToNewSession()` in `src/session-runtime/connect-load.ts` only handles `-32603` (Internal Error) for the fallback-to-new-session logic, but does **not** handle `-32602`
4. The `-32602` error propagates up uncaught, causing the process to exit with code 1

## Expected Behavior

When `session/load` returns `-32602` (Invalid params) or `-32601` (Method not found), acpx should gracefully fallback to `session/new` instead of crashing.

## Suggested Fix

The constant `SESSION_CONTROL_UNSUPPORTED_ACP_CODES` (containing `-32601` and `-32602`) already exists in the codebase but is not used in `shouldFallbackToNewSession`. The fix is to check it **before** the `sessionHasAgentMessages` guard:

```diff
 function shouldFallbackToNewSession(error, record) {
   if (error instanceof TimeoutError || error instanceof InterruptedError) return false;
   if (isAcpResourceNotFoundError(error)) return true;
+  const acpErr = extractAcpError(error);
+  if (acpErr && SESSION_CONTROL_UNSUPPORTED_ACP_CODES.has(acpErr.code)) return true;
   if (!sessionHasAgentMessages(record)) {
     if (isAcpQueryClosedBeforeResponseError(error)) return true;
-    if (extractAcpError(error)?.code === -32603) return true;
+    if (acpErr?.code === -32603) return true;
   }
   return false;
 }
```

### Why this works

- `SESSION_CONTROL_UNSUPPORTED_ACP_CODES = new Set([-32601, -32602])` already exists in the codebase
- `-32601` = JSON-RPC Method not found (agent doesn't support the method at all)
- `-32602` = JSON-RPC Invalid params (method exists but params validation fails — this is what cursor-agent returns for `session/load`)
- Checking these codes **unconditionally** (not gated behind `sessionHasAgentMessages`) ensures that even sessions with prior agent messages can gracefully recover when the agent simply doesn't support `session/load`
- As a bonus, the `extractAcpError(error)` call is now cached in a local variable to avoid redundant extraction

## Steps to Reproduce

1. Configure acpx with cursor-agent as the backend
2. Run `acpx cursor exec "hello"` — **works** (uses `session/new`)
3. Run `acpx cursor prompt` and send a message — **works** (first run, creates new session)
4. Run `acpx cursor prompt` again — **crashes** (tries `session/load` on existing session, gets `-32602`)

## Workaround

Manually patch the compiled `dist/session-*.js` file to add the `SESSION_CONTROL_UNSUPPORTED_ACP_CODES` check, or delete `~/.acpx/sessions/*` before each run (forces fresh sessions every time).

## Environment

- acpx: 0.3.0
- cursor-agent (Cursor CLI): 2026.03.11-6dfa30c
- OS: Windows 10
- Node.js: v24.14.0
- Integration: OpenClaw gateway calling acpx in prompt mode

Bortlesboat:
## Summary

- `shouldFallbackToNewSession()` now handles JSON-RPC `-32601` (Method not found) and `-32602` (Invalid params) unconditionally, falling back to `session/new` instead of crashing
- These codes indicate the agent doesn't properly support `session/load` (e.g., cursor-agent returns `-32602`), so fallback should happen regardless of whether the session has prior agent messages
- The `extractAcpError()` call is hoisted above the `sessionHasAgentMessages` guard to cache the result and avoid redundant extraction

## Test plan

- [x] Added test: `-32602` with agent messages triggers fallback (was crashing before)
- [x] Added test: `-32601` with agent messages triggers fallback
- [x] All 8 connect-load tests pass
- [x] Existing `-32603` behavior unchanged (still gated behind `sessionHasAgentMessages`)
- [x] `tsc --noEmit`, `oxlint`, `oxfmt` clean

Closes #152

### DIFF
--- a/src/session-runtime/connect-load.ts
+++ b/src/session-runtime/connect-load.ts
@@ -36,12 +36,21 @@ export type ConnectAndLoadSessionResult = {
   loadError?: string;
 };
 
+// JSON-RPC codes that indicate the agent does not support session/load.
+// -32601 = Method not found, -32602 = Invalid params.
+const SESSION_LOAD_UNSUPPORTED_CODES = new Set([-32601, -32602]);
+
 function shouldFallbackToNewSession(error: unknown, record: SessionRecord): boolean {
   if (error instanceof TimeoutError || error instanceof InterruptedError) {
     return false;
   }
 
   if (isAcpResourceNotFoundError(error)) {
+    return true;
+  }
+
+  const acp = extractAcpError(error);
+  if (acp && SESSION_LOAD_UNSUPPORTED_CODES.has(acp.code)) {
     return true;
   }
 
@@ -52,7 +61,6 @@ function shouldFallbackToNewSession(error: unknown, record: SessionRecord): bool
       return true;
     }
 
-    const acp = extractAcpError(error);
     if (acp?.code === -32603) {
       return true;
     }
--- a/test/connect-load.test.ts
+++ b/test/connect-load.test.ts
@@ -215,6 +215,114 @@ test("connectAndLoadSession falls back to createSession for empty sessions on ad
   });
 });
 
+test("connectAndLoadSession falls back to session/new on -32602 Invalid params", async () => {
+  await withTempHome(async (homeDir) => {
+    const cwd = path.join(homeDir, "workspace");
+    await fs.mkdir(cwd, { recursive: true });
+
+    const record = makeSessionRecord({
+      acpxRecordId: "invalid-params-record",
+      acpSessionId: "invalid-params-session",
+      agentCommand: "agent",
+      cwd,
+      messages: [
+        {
+          Agent: {
+            content: [{ Text: "has history" }],
+            tool_results: {},
+          },
+        },
+      ],
+    });
+
+    const client: FakeClient = {
+      hasReusableSession: () => false,
+      start: async () => {},
+      getAgentLifecycleSnapshot: () => ({
+        running: true,
+      }),
+      supportsLoadSession: () => true,
+      loadSessionWithOptions: async () => {
+        throw {
+          error: {
+            code: -32602,
+            message: "Invalid params",
+          },
+        };
+      },
+      createSession: async () => ({
+        sessionId: "fallback-from-32602",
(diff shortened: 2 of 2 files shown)


osolmaz:
## Triage result

### Quick read
- Intent valid: ✅ Yes
- Solves the right problem: ✅ Yes
- Validation: ✅ Bug reproduced and fixed
- Close PR: ✅ No
- Refactor needed: ✅ None
- Human decision needed: ready for human landing decision

### Intent
> Make `acpx` recover cleanly when an ACP backend cannot actually restore a saved session, so reopening a persistent session starts a fresh ACP session instead of crashing.

### Why
- The change targets the actual fallback gate that was dropping `-32601` and `-32602` load failures on the floor.
- The reported failure mode was reproduced locally by ablating the new fallback logic, then the restored PR fix changed the outcome back to passing.
- The patch is narrow, the added tests cover the claimed crash path, and no refactor is needed first.

### Codex review
- Status: ✅ Clear
- Notes: No existing GitHub Codex review findings were present. Fresh local review against the refreshed merge base did not surface a blocking P0/P1 issue before Codex transport/auth interruptions.

### CI/CD
- Status: 🚦 Green
- Notes: The fork PR CI run was initially waiting on approval. I approved it, reran it, and the current workflow/checks for head `831d8e0` completed successfully.

### Recommendation
🏁 escalate to a human

PR #207, opened 2026-04-01
fix: Improve uniqueness of queue owner generation IDs

Yuan-ManX:


steipete:
Landed via maintainer rewrite and rebase onto main.

- Gate: `pnpm run check` passed locally; GitHub checks are green on the rewritten head
- Targeted proof: added a queue lease regression that pins `Date.now()` and `Math.random()` so the previous generation formula collides deterministically; the new `crypto.randomInt()`-backed generation stays distinct, positive, and safe-integer
- Land commit: https://github.com/openclaw/acpx/commit/3a0007e99d22c4fe514184f84fa54cdf2145b66d
- Source rewrite commit before rebase: https://github.com/Yuan-ManX/acpx/commit/8eb43fb393495830db4f6b6eefe45f0f17db177c

Thanks @Yuan-ManX!

PR #208, opened 2026-04-01
fix: Improve writeSessionRecord for Windows platform compatibility

Yuan-ManX:


steipete:
I’m not landing this as-is.

`writeSessionRecord` currently writes a temp file and then uses `fs.rename(tempFile, file)`. Node’s `fs.rename` contract says an existing file destination is overwritten: https://nodejs.org/api/fs.html#fsrenameoldpath-newpath-callback

The proposed Windows branch unlinks the current session file before the rename. That removes the useful atomic-replace property and creates a window where a crash leaves no session file at all.

A safer follow-up would need a concrete Windows repro for the current code path plus a fix that preserves atomic replacement semantics, or a platform-specific fallback only for the exact error we can reproduce.

PR #216, opened 2026-04-04
fix: restrict IPC socket directory permissions to owner-only

garagon:
The queue socket directory at `/tmp/acpx-<hash>/` and the queue base directory at `~/.acpx/queues/` are created with `fs.mkdir` without an explicit `mode`, so they inherit the default umask (typically `0o755`). This allows other local users to traverse the directory and discover session socket files.

This change sets `mode: 0o700` on both directories so only the owning user can access the IPC sockets.

**What changed:**
- `ensureQueueDir()` in `src/queue-lease-store.ts` now passes `mode: 0o700` to both `mkdir` calls

**What did NOT change:**
- No changes to the IPC protocol or queue behavior
- No changes to socket file creation (Unix sockets inherit directory permissions for access control)

**How I verified:**
- Reviewed that `queueBaseDir()` resolves to `~/.acpx/queues/` and `queueSocketBaseDir()` resolves to `/tmp/acpx-<hash>/`
- Confirmed no other code path creates these directories
- Confirmed `mode` is respected by Node.js `fs.mkdir` on Linux and macOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Repo: https://github.com/openclaw/acpx
 ### Fixes
 
 - Runtime/persistent sessions: keep reusable persistent ACP clients warm across turns and close pooled clients during runtime close. (#265) Thanks @Sway-Chan.
+- CLI/queue: tighten persistent queue and IPC socket directories to owner-only permissions, including previously-created permissive directories. (#216) Thanks @garagon.
 - Runtime/ACP: drain late post-success session updates before closing prompt turns so adapters that resolve `session/prompt` before final updates do not drop assistant output. (#251) Thanks @logofet85-ai.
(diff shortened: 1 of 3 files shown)


steipete:
Landed via maintainer fixup and rebase onto main.

- Gate: `pnpm run check` passed locally after rebase (format, typecheck, lint, build, viewer type/build, coverage: 534 tests)
- Targeted proof: queue lease regression pre-creates both queue directories as `0777`, then `tryAcquireQueueOwnerLease` tightens `~/.acpx/queues` and `/tmp/acpx-<hash>` to `0700`
- Land commit: https://github.com/openclaw/acpx/commit/b201fd13bb7d3bef380772fc54937842b8e75211
- Source fixup commit before rebase: https://github.com/garagon/acpx/commit/0c51e7755d1c21c7739ccdf24646d9399d50b0d8

Thanks @garagon!

Issue #206, opened 2026-04-01
Feature:  command to bulk-delete closed sessions


whtoo:
## Problem

Currently, `acpx kimi sessions close <name>` only marks a session as \`[closed]\` in the sessions list but does **not** delete the session files from \`~/.acpx/sessions/\`. Over time this leads to:

- Hundreds of orphaned session files accumulating on disk (in our case: 190 closed sessions = ~2.4GB)
- The sessions list shows a cluttered mix of open and closed entries
- No way to bulk-clean closed sessions without manual file deletion

## Proposed Solution

Add a \`prune\` subcommand to \`acpx <agent> sessions\`:

\`\`\`bash
# Prune all closed sessions for current agent/cwd
acpx kimi sessions prune

# Dry-run: show what would be deleted
acpx kimi sessions prune --dry-run

# Prune sessions closed before a given date
acpx kimi sessions prune --before 2026-03-01

# Prune sessions older than N days
acpx kimi sessions prune --older-than 7

# Also delete session history files (.stream.ndjson)
acpx kimi sessions prune --include-history
\`\`\`

## Behavior

1. Query the sessions list for the current agent
2. Filter sessions with status = \`[closed]\`
3. Apply optional date/age filters
4. Delete:
   - \`{session_id}.json\` (session metadata)
   - \`{session_id}.stream.ndjson\` and \`{session_id}.stream.N.ndjson\` (if \`--include-history\`)
5. Print summary: X sessions pruned, Y MB freed

## Motivation

- **Disk hygiene**: sessions can grow large (hundreds of MB per session with long conversations)
- **UX**: keeps \`sessions list\` output clean and actionable
- **Alternative**: implementing file deletion inside \`sessions close\` is also acceptable, but a separate \`prune\` command gives users more control

## Alternatives Considered

| Approach | Pros | Cons |
|----------|------|------|
| \`sessions close\` also deletes files | Simple, no new command | Breaking change if users rely on file persistence |
| \`sessions prune\` (proposed) | Granular control, dry-run possible | Slightly more complex |
| Automatic pruning on close | Zero maintenance | May delete data users want to keep |

This is purely a CLI UX improvement; the session state is already tracked in memory so no protocol changes needed.

steipete:
Fixed on current `main` by #227 / commit 0ad1577.

The implemented command is `acpx <agent> sessions prune` with:
- `--dry-run`
- `--before <date>`
- `--older-than <days>`
- `--include-history`

[...]

PR #227, opened 2026-04-06
feat(sessions): add prune command to delete closed sessions

coder999999999:
## Summary

- Adds `acpx sessions prune` to delete closed sessions and free disk space
- Supports `--dry-run` to preview what would be removed without deleting anything
- Supports `--before <date>` and `--older-than <days>` to filter by age (`--before` takes precedence if both are supplied)
- Supports `--include-history` to also delete event stream files (`.stream.ndjson`)
- Output works in `text`, `json`, and `quiet` formats

## Changes

- `src/session/persistence/repository.ts` — core `pruneSessions` function; filters closed sessions from the index, deletes their JSON records, optionally deletes stream files, then rebuilds the index
- `src/cli/flags.ts` — `SessionsPruneFlags` type and parsers for `--before` (returns `Date`) and `--older-than`
- `src/cli/command-handlers.ts` — `handleSessionsPrune` handler
- `src/cli/command-registration.ts` — registers the `prune` subcommand under `sessions`
- `src/cli/output/render.ts` — `printPruneResultByFormat` with text, json, and quiet output
- `test/sessions-prune.test.ts` — 8 tests covering all flag combinations and edge cases

## Test plan

- [ ] `sessions prune --dry-run` lists sessions without deleting
- [ ] `sessions prune` deletes closed sessions and reports bytes freed
- [ ] `sessions prune --before 2026-01-01` respects date cutoff
- [ ] `sessions prune --older-than 30` respects day threshold
- [ ] `sessions prune --include-history` removes stream files too
- [ ] `sessions prune --format json` emits valid JSON
- [ ] Open sessions are never touched

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Repo: https://github.com/openclaw/acpx
 ### Changes
 
 - CLI/claude: add `--system-prompt <text>` and `--append-system-prompt <text>` global flags that forward through ACP `_meta.systemPrompt` on `session/new`, letting callers replace or append to the Claude Code system prompt without dropping out of persistent acpx sessions. The value is persisted in `session_options.system_prompt` so ensure/reuse flows keep the override. Codex and other agents ignore the field. (#229) Thanks @Vercantez.
+- CLI/sessions: add `sessions prune` with `--dry-run`, age filters, and `--include-history` so closed session records and optional event streams can be cleaned up explicitly. (#227) Thanks @coder999999999.
 - Runtime/embedding: add `startTurn(...)` turn handles so embedders can observe live runtime events separately from terminal completion, cancel a turn, or close only the event stream while preserving `runTurn(...)` compatibility. (#262) Thanks @enki.
 
 ### Breaking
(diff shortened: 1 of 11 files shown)


steipete:
Thanks, this is a real UX gap and #206 is worth fixing, but I would not land this PR as-is yet.

Main blockers from review:

1. The new behavior has no CLI coverage. The tests call `pruneSessions()` directly, so they do not prove command registration, flag parsing, default/explicit agent scoping, or text/json/quiet output. Please add at least CLI-level coverage for `acpx <agent> sessions prune --dry-run`, `--format json`, and `--include-history`.

2. The age filters are described as pruning sessions closed before / older than a cutoff, but the implementation filters on `lastUsedAt`. `closeSession()` currently sets `lastUsedAt = closedAt`, but restored/imported/older records can have those diverge. Either filter on `closedAt ?? lastUsedAt` and test that, or change the CLI wording/tests to say last activity.

3. This is user-facing CLI surface, so it needs a changelog entry and a short docs/help sync. The issue specifically asks for disk hygiene around closed sessions; users should be able to discover the safe dry-run path.

The storage-side shape is directionally fine, and keeping deletion explicit behind `prune` instead of changing `sessions close` behavior seems right. I’d re-review after the above is covered.

PR #230, opened 2026-04-08
feat(agent-registry): prefer globally installed agent binaries over npx

ApolloS-Hub:
## Summary

When a built-in agent adapter (e.g. `claude-agent-acp`, `codex-acp`) is globally installed and available on `PATH`, use it directly instead of falling back to `npx` package-exec.

## Problem

The current agent resolution only checks:
1. Local `node_modules` install
2. `npx` package-exec bridge

In slow-network environments (high latency to npm registry, corporate proxies, China mainland, restricted networks), the `npx` download/cache-check during probe can time out, causing the acpx backend to stay at `registered` and never reach `ready`. This manifests as `sessions_spawn(runtime="acp")` hanging indefinitely.

**Related: openclaw/openclaw#51345** — `sessions_spawn(runtime="acp") hangs immediately`

## Solution

Add a new resolution step between local install and npx fallback:

```
1. Local node_modules install (existing)
2. Global PATH binary (NEW — via which/where)  ← added
3. npx package-exec bridge (existing fallback)
```

New `resolveGlobalPathAgentLaunch()` function uses `which` (Unix) or `where` (Windows) with a 5-second timeout to detect globally installed binaries.

## Changes

- **`src/agent-registry.ts`**: Added `resolveGlobalPathAgentLaunch()` and integrated it into `resolveBuiltInAgentLaunch()` resolution chain
- **`src/acp/client.ts`**: Added spawn log message for `global-path` source
- **`test/agent-registry.test.ts`**: Added 4 new tests covering the global PATH resolution

## Benefits

- Eliminates network dependency during agent spawn for users who pre-install adapters
- Faster startup in all environments (no npx cache check overhead)
- Zero breaking changes — existing behavior preserved as fallback
- Cross-platform: works on Unix (`which`) and Windows (`where`)

## Test Results

All 16 tests pass, including 4 new tests:
- `resolveGlobalPathAgentLaunch uses a globally installed binary from PATH`
- `resolveGlobalPathAgentLaunch returns undefined when binary is not on PATH`
- `resolveGlobalPathAgentLaunch ignores non-built-in commands`
- `resolveBuiltInAgentLaunch prefers global PATH over package-exec when local install unavailable`

### DIFF
--- a/src/acp/client.ts
+++ b/src/acp/client.ts
@@ -417,6 +417,10 @@ export class AcpClient {
       this.log(
         `spawning installed built-in agent ${resolvedBuiltInLaunch.packageName}${resolvedBuiltInLaunch.packageVersion ? `@${resolvedBuiltInLaunch.packageVersion}` : ""} via ${spawnCommand} ${args.join(" ")}`,
       );
+    } else if (resolvedBuiltInLaunch?.source === "global-path") {
+      this.log(
+        `spawning globally installed built-in agent ${resolvedBuiltInLaunch.packageName} via PATH binary ${spawnCommand}`,
+      );
     } else if (resolvedBuiltInLaunch?.source === "package-exec") {
       this.log(
         `spawning built-in agent ${resolvedBuiltInLaunch.packageName}@${resolvedBuiltInLaunch.packageRange} via current Node package exec bridge ${spawnCommand} ${args.join(" ")}`,
--- a/src/agent-registry.ts
+++ b/src/agent-registry.ts
@@ -1,3 +1,4 @@
+import { execFileSync } from "node:child_process";
 import fs from "node:fs";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
@@ -17,7 +18,7 @@ type BuiltInAgentPackageSpec = {
 };
 
 type BuiltInAgentLaunch = {
-  source: "installed" | "package-exec";
+  source: "installed" | "global-path" | "package-exec";
   command: string;
   args: string[];
   packageName: string;
@@ -33,6 +34,7 @@ type BuiltInLaunchResolverOptions = {
   resolvePackageRoot?: (packageName: string) => string;
   execPath?: string;
   resolveNpmCliPath?: (execPath: string) => string;
+  resolveGlobalBinPath?: (binName: string) => string | undefined;
 };
 
 export const AGENT_REGISTRY: Record<string, string> = {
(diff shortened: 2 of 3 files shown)


osolmaz:
## Triage result
### Quick read
- Human decision needed: ready for human landing decision
- Verdict: `good_enough` bug fix is validated, review is clear, CI is green, and merge conflict checks are clean.

### Intent
Make `acpx` prefer globally installed built-in agent binaries on `PATH` before falling back to `npx`, improving reliability and startup speed.

### Why
When a built-in adapter is not locally installed, the previous resolution path jumped to `npx`; in slow/restricted networks this could stall and leave ACP startup stuck at `registered` instead of `ready`. This PR adds a `global-path` step between installed and package-exec while preserving existing fallbacks.

### Codex review
- GitHub Codex review findings for this head: none.
- Local Codex review: completed successfully with substantive output and no actionable bugs.
- Focused validation confirmed behavior change:
  - PR code passes targeted check (`resolveBuiltInAgentLaunch` returns `source === "global-path"` when local install is unavailable).
  - Temporary ablation of the new fallback hop causes failure (`actual: 'package-exec'`, `expected: 'global-path'`).
  - Restoring PR code returns the targeted check to passing.

### CI/CD
- CI was initially approval-blocked (`action_required`).
- Workflow approval was attempted and succeeded.
- The approved rerun was monitored to completion and finished `success` (run attempt 2).
- Initial and final merge-conflict checks against `origin/main`: clean.

### Recommendation
Recommendation: 🏁 escalate to a human.
This PR appears technically ready to land and now needs a human landing decision.

ApolloS-Hub:
Friendly bump 👋 — CI is green, triage review was completed with a `good_enough` verdict, and no merge conflicts. Is there anything else needed to get this landed?

steipete:
Codex review: I would not land this default behavior as-is.

The underlying pain is real, but this changes the built-in launcher contract from an acpx-owned pinned adapter (`@zed-industries/codex-acp@...`, `@agentclientprotocol/claude-agent-acp@...`) to whichever matching binary happens to be first on `PATH`. That can silently pick an old or incompatible global adapter and bypass the version range acpx owns.

This also conflicts with the design note in `docs/2026-04-06-built-in-agent-launch-ownership.md`: built-ins should keep package/version/launch strategy owned by acpx; users or embeddings may provide explicit command overrides when they want a custom/global binary. The issue workaround already has that explicit override shape, which is safer because the user intentionally opts into the global binary.

A safer direction would be one of:
- keep global binaries as explicit `agents.<name>.command` config only, and improve docs/error text for slow `npx`/probe environments; or
- add an acpx-owned adapter materialization/cache path that installs the pinned adapter version, then launches it directly with the current Node runtime.

So: useful report, but automatic PATH preference is too implicit for built-in defaults.

PR #248, opened 2026-04-15
fix: require explicit ACPX auth envs for ACP auth selection

### LINKED ISSUE #247: acpx can overwrite shared Codex CLI auth and switch Codex from OAuth to OPENAI_API_KEY auth
## Summary

`openclaw/acpx` shares the same Codex CLI auth state that the normal `codex` CLI uses. That is expected.

The bug is that starting an `acpx` Codex session can inadvertently overwrite that shared auth state. If the parent environment has `OPENAI_API_KEY` set, `acpx` may authenticate Codex through the API-key path and replace the existing Codex OAuth auth on disk. After that, normal `codex` CLI sessions stop using ChatGPT OAuth and start using the API key instead.

## What is going wrong

The shared auth state is being treated as writable session startup state, not just reusable auth state.

In practice, this means:

- Codex CLI is already logged in with ChatGPT OAuth
- another tool starts a Codex session through `acpx`
- `acpx` sees `OPENAI_API_KEY` in the environment
- `acpx` authenticates Codex via the API-key method
- the existing Codex auth is overwritten
- from that point on, `codex` CLI uses API-key auth instead of the previous OAuth login

## Expected behavior

If Codex CLI auth is shared, `acpx` should be able to reuse it without replacing it.

Ambient env vars like `OPENAI_API_KEY` should not cause `acpx` to rewrite an existing Codex OAuth login. If ACP auth is meant to happen explicitly, it should only happen from explicit ACP auth configuration.

## Actual behavior

`acpx` session startup can replace an existing Codex OAuth login with API-key auth when `OPENAI_API_KEY` is present in the environment.

This is persistent because it changes the shared Codex auth state on disk.

## Reproduction

1. Log into the normal `codex` CLI with ChatGPT OAuth.
2. Confirm Codex is using OAuth auth.
3. In the parent environment, set:
   ```bash
   export OPENAI_API_KEY=sk-...
   ```
4. Start a Codex session through `acpx`:
   ```bash
   acpx codex sessions new
   ```
5. Check Codex auth again.

## Result

The existing Codex OAuth auth can be replaced, and the normal `codex` CLI begins using the API key instead.

## Why this matters

This makes `acpx` session startup destructive to shared Codex auth state. The user can do the right thing once by logging into Codex with OAuth, then lose that auth simply because an `acpx` session was started in an environment that happened to include `OPENAI_API_KEY`.

## Likely fix

`acpx` should separate:

- shared Codex auth reuse
- explicit ACP auth negotiation

Ambient provider env vars like `OPENAI_API_KEY` can still be passed through to child processes, but they should not by themselves cause `acpx` to overwrite existing shared Codex auth.

valkyriweb:
Fixes #247

`acpx` and the normal `codex` CLI intentionally share the same Codex auth state. That part is fine.

The bug is that starting a Codex session through `acpx` could overwrite that shared auth instead of just reusing it.

If the parent environment had `OPENAI_API_KEY` set, `acpx` could treat it as an ACP auth credential, select `openai-api-key` during startup, and send `codex-acp` down its API-key login path. In practice that meant an existing ChatGPT OAuth login could be replaced on disk, and later `codex` CLI runs would start using the API key instead.

This change keeps explicit ACP auth working, but stops ambient provider env from participating in auth-method selection.

Only these should drive ACP `authenticate` selection:

- config `auth` entries
- explicit `ACPX_AUTH_<METHOD_ID>` env vars

Ambient provider env like `OPENAI_API_KEY` can still exist in the process environment, but it should not be enough on its own to rewrite shared Codex auth.

## What changed

- only `ACPX_AUTH_*` env vars count as ACP auth input during auth-method selection
- config `auth` entries still work as before
- explicit `ACPX_AUTH_*` values are still promoted into the child environment for adapters that expect normalized names like `OPENAI_API_KEY`
- docs were updated to make the explicit-auth requirement clear
- regression tests were added for both auth selection and child env propagation

## Verification

- `pnpm run check`
- `pnpm run check:docs`

_AI-assisted change._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Repo: https://github.com/openclaw/acpx
 - Sessions/load: fall back to a fresh ACP session when adapters reject `session/load` with JSON-RPC `-32601` or `-32602`, so persistent session reconnects do not crash on partial load support. (#174) Thanks @Bortlesboat.
 - Flows/runtime: finalize interrupted `flow run` bundles as failed instead of leaving them stuck at `running` when the process receives `SIGHUP`, `SIGINT`, or `SIGTERM`.
 - Client/auth: cache derived auth env key lists per auth method to avoid repeated allocations during credential lookup. (#167) Thanks @Yuan-ManX.
+- Client/auth: require explicit `ACPX_AUTH_*` env vars or config `auth` entries for ACP auth-method selection, so ambient provider env like `OPENAI_API_KEY` no longer triggers unintended login flows in adapters such as `codex-acp`.
 
 ## 2026.3.12 (v0.3.0)
 
--- a/README.md
+++ b/README.md
@@ -281,6 +281,11 @@ Supported keys:
 
 Use `acpx config show` to inspect the resolved result and `acpx config init` to create the global template.
 
+For ACP `authenticate` handshakes, use either config `auth` entries or explicit
+`ACPX_AUTH_<METHOD_ID>` environment variables such as `ACPX_AUTH_OPENAI_API_KEY`.
+Ambient provider env vars such as `OPENAI_API_KEY` are still passed through to
+child agents, but they do not trigger ACP auth-method selection on their own.
+
 ## Output formats
 
 ```bash
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -380,6 +380,11 @@ Supported keys:
 
 CLI flags always override config values.
 
+For ACP `authenticate` handshakes, use either config `auth` entries or explicit
+`ACPX_AUTH_<METHOD_ID>` environment variables such as `ACPX_AUTH_OPENAI_API_KEY`.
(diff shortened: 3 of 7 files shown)


steipete:
Landed via maintainer rebase onto main.

- Gate: `pnpm run check` passed locally after rebase (format, typecheck, lint, build, viewer type/build, coverage: 533 tests)
- Targeted proof: auth tests cover explicit `ACPX_AUTH_*` selection, ambient `OPENAI_API_KEY` being ignored for ACP auth-method selection, and explicit ACPX env promotion into the child agent environment
- Land commit: https://github.com/openclaw/acpx/commit/351f1588c169b096d797521fc27da37da4998b64
- Source commit before rebase: https://github.com/valkyriweb/acpx/commit/71f5b70956a70e9bfe1e74a5f5ffafec28c4bed1

Thanks @valkyriweb!

PR #249, opened 2026-04-16
feat: forward proposedSessionId in session/new via createSession hints

henkterharmsel:
## What

Three related fixes for reliable ACP session resume in \`acpx\`:

### 1. \`createSession(cwd, hints?)\` — forward \`proposedSessionId\`

\`createSession\` now accepts an optional \`hints\` object. When \`hints.proposedSessionId\` is provided, it is included in the \`_meta\` field of the \`session/new\` ACP request.

All internal \`createSession\` call sites in the runtime engine (\`manager.ts\` and \`reconnect.ts\`) now derive the session UUID from the ACP session key (\`agent:claude:acp:<UUID>\` → \`.split(':').pop()\`) and pass it as \`proposedSessionId\`. This gives Claude Code sessions a stable, predictable file name — required for reliable resume.

### 2. Oneshot resume double-load fix

\`ensureSession\` (phase 1) previously started Claude, loaded the session, then shut it down via \`client.close()\`. This wrote a \`closed\` marker to the JSONL file. \`connectAndLoadSession\` (phase 2) then failed because the JSONL was already closed.

For oneshot sessions with a \`resumeSessionId\`, \`ensureSession\` now skips the start/load/close cycle entirely, constructs a minimal session record, and returns early. Phase 2 performs the actual resume.

### 3. Oneshot second-turn implicit resume

When a follow-up message arrives for a completed oneshot session (e.g. via a Telegram reply after Tommy finished a task), the caller does not provide \`resumeSessionId\` because \`persistedResumeSessionId\` is \`undefined\` in oneshot mode. The normal path then calls \`createSession\` with the same \`proposedSessionId\` — but the JSONL already exists from the first turn, causing Claude Code to fail with a session-init error (\`ACP_SESSION_INIT_FAILED\`).

Fix: if the existing record has an \`acpSessionId\` (written during the first turn), \`ensureSession\` now uses it as an implicit resume — same early-return pattern as fix #2 above, without requiring an explicit \`resumeSessionId\` hint from the caller.

## Why

Without these fixes, resuming a completed oneshot session in OpenClaw returns "Internal error". New sessions work fine; it is specifically resume (and second turns) that fail.

Root causes:
1. \`claude-agent-acp\` used \`randomUUID()\` for every session, making file names unpredictable.
2. \`acpx\` did not support passing \`proposedSessionId\` through \`createSession\` or its internal call sites.
3. \`ensureSession\`'s two-phase design closed the JSONL before phase 2 could use it.
4. Second turns via \`dispatch-acp\` called \`ensureSession\` without \`resumeSessionId\`, triggering a session-init conflict with the existing JSONL.

## Backward Compatibility

All changes are additive. Existing behavior is preserved for non-oneshot modes and sessions without an existing record. Passing \`proposedSessionId\` is opt-in.

## Dependencies

This PR is part of a two-repo fix for ACP session resume. Merge order:

1. \`agentclientprotocol/claude-agent-acp#554\` — support \`proposedSessionId\` in \`_meta\` and fix UUID extraction *(merge first)*
2. \`openclaw/acpx\` — this PR *(merge after #1)*

Note: an openclaw PR is not required. OpenClaw already passes the correct \`resumeSessionId\` and \`sessionKey\` parameters to \`ensureSession\` — all fixes are encapsulated in \`acpx\`.

Each PR is backward-compatible with the previous version of its dependency.

## Testing

- New tests added for \`proposedSessionId\` forwarding (in \`createSession\` and reconnect fallback paths), oneshot resume bypass, UUID extraction, and second-turn implicit resume.
- Full test suite passes: \`pnpm test\` (527/527)
- Build and lint clean: \`pnpm build && pnpm check\`
- End-to-end verified locally: OpenClaw oneshot sessions resume correctly with multi-turn history preserved (OpenClaw 2026.4.12 + acpx 0.5.3 + claude-agent-acp 0.27.0).

## AI Assistance 🤖

Developed with AI assistance (Claude Code). Logic derived from locally-applied patches confirmed working. Author (Henk ter Harmsel) has reviewed and understands all changes.

### DIFF
--- a/src/acp/client.ts
+++ b/src/acp/client.ts
@@ -622,10 +622,30 @@ export class AcpClient {
     return { readable, writable };
   }
 
-  async createSession(cwd = this.options.cwd): Promise<SessionCreateResult> {
+  async createSession(
+    cwd = this.options.cwd,
+    hints?: { proposedSessionId?: string },
+  ): Promise<SessionCreateResult> {
     const connection = this.getConnection();
     const { command, args } = splitCommandLine(this.options.agentCommand);
     const claudeAcp = isClaudeAcpCommand(command, args);
+
+    let sessionMeta = buildClaudeCodeOptionsMeta(this.options.sessionOptions);
+    if (hints?.proposedSessionId) {
+      sessionMeta = sessionMeta ?? {};
+      sessionMeta = {
+        ...sessionMeta,
+        claudeCode: {
+          ...(sessionMeta.claudeCode as Record<string, unknown> | undefined),
+          options: {
+            ...((sessionMeta.claudeCode as Record<string, unknown> | undefined)?.options as
+              | Record<string, unknown>
+              | undefined),
+            proposedSessionId: hints.proposedSessionId,
+          },
+        },
+      };
+    }
 
     let result: Awaited<ReturnType<typeof connection.newSession>>;
     try {
@@ -633,7 +653,7 @@ export class AcpClient {
         connection.newSession({
           cwd: asAbsoluteCwd(cwd),
           mcpServers: this.options.mcpServers ?? [],
-          _meta: buildClaudeCodeOptionsMeta(this.options.sessionOptions),
+          _meta: sessionMeta,
         }),
       );
       result = claudeAcp
--- a/src/runtime/engine/manager.ts
+++ b/src/runtime/engine/manager.ts
@@ -273,6 +273,26 @@ export class AcpRuntimeManager {
       return record;
     }
 
+    // For oneshot second turns arriving without an explicit resumeSessionId (e.g. a follow-up
+    // message via dispatch-acp after a completed turn), the caller does not know to pass
+    // resumeSessionId because persistedResumeSessionId is undefined in oneshot mode.
+    // The normal path would call createSession(cwd, { proposedSessionId: UUID }) but the JSONL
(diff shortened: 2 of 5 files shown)


steipete:
Codex review: not landing this as-is yet.

Two blockers from review/repro:

1. The declared upstream dependency, https://github.com/agentclientprotocol/claude-agent-acp/pull/554, is still open. This PR depends on that adapter accepting `_meta.claudeCode.options.proposedSessionId`, so merging acpx first would ship a forwarder before the receiving side is actually available upstream.

2. The implicit oneshot second-turn path does not appear to work with the current store keying. `ensureSession()` does `sessionStore.load(input.sessionKey)`, but oneshot records are saved under `createRecordId(input.sessionKey, "oneshot")`, i.e. `sessionKey:oneshot:<randomUUID>`. I reproduced this on the PR branch with an existing oneshot record under `agent:claude:acp:<uuid>:oneshot:<uuid>`: `ensureSession({ sessionKey, mode: "oneshot" })` called `start()` and `createSession()` and returned a new `fresh-backend-session`, leaving the existing backend session unused.

The minimal next step is a failing regression test for that exact store shape, then fix the lookup strategy or API contract so a follow-up oneshot can actually identify the previous acpx record. After upstream claude-agent-acp lands, the `proposedSessionId` forwarding part can be reconsidered separately.

PR #256, opened 2026-04-17
fix: add text remediation hints for ACP session errors

SJeffZhang:
## Summary

This is a small first pass for #176.

It keeps the JSON / machine-readable error contract unchanged and adds additive remediation hints only to text-mode error output.

Initial text-mode coverage in this pass:

- `AUTH_REQUIRED`
- `NO_SESSION`
- common ACP runtime/session-control failures such as `session/load`, `session/set_mode`, `session/set_model`, and `session/set_config_option`

## Why

`acpx` already normalizes these failures into stable `code`, `detailCode`, `origin`, and raw `acp` metadata, but text mode usually stopped at the raw error message. That left users without a clear next step even when the client already had enough information to point them in the right direction.

## User Impact

Text-mode CLI failures now include short next-step hints in a few high-value cases, for example:

- add the missing auth credential under `auth.<method>`
- start a fresh session with `acpx <agent> sessions new`
- rerun with `--verbose` to capture ACP method/error details

JSON output and machine-readable error fields stay unchanged.

## Root Cause

The existing text formatter and top-level stderr path emitted normalized error messages, but they did not translate the existing normalized metadata into actionable remediation guidance for human readers.

## Validation

Targeted local validation:

- `corepack pnpm run build:test`
- `node --test --test-name-pattern 'normalizeOutputError maps ACP resource not found errors to NO_SESSION|normalizeOutputError maps AuthPolicyError to AUTH_REQUIRED detail|normalizeOutputError infers AUTH_REQUIRED detail from ACP payload|sessions new --resume-session fails when agent does not support session/load|sessions new --resume-session surfaces not-found loadSession errors without fallback|set-mode surfaces actionable guidance when agent rejects session/set_mode params|prompt exits with NO_SESSION when no session exists \(no auto-create\)' dist-test/test/error-normalization.test.js dist-test/test/cli.test.js`
- local `node --input-type=module -e ...` assertions for generated remediation hints
- `corepack pnpm run typecheck`
- `corepack pnpm run format:check`
- `oxlint --type-aware src`
- `./node_modules/.bin/tsx scripts/lint-persisted-key-casing.ts`
- `./node_modules/.bin/tsx scripts/lint-flow-schema-terms.ts`

Manual text-mode end-to-end stderr checks:

- `AUTH_REQUIRED`
- failed `session/load` / `NO_SESSION`
- `session/set_mode` invalid params

## AI Assistance

AI-assisted with Codex. I reviewed the touched code paths and local validation results directly.

PR #261, opened 2026-04-21
Surface actionable prompt failure diagnostics

enki:
## Summary

This adds actionable context when a prompt request fails with a generic ACP internal error.

- Decorates generic prompt failures with ACP error payload details when available.
- Falls back to a bounded stderr tail when the ACP payload does not include a useful message.
- Preserves the original error as the cause.
- Adds client tests for payload-detail and stderr-tail diagnostics.

## Validation

- `pnpm run build:test`
- `node --test dist-test/test/client.test.js`

### DIFF
--- a/src/acp/client.ts
+++ b/src/acp/client.ts
@@ -104,6 +104,7 @@ const DRAIN_POLL_INTERVAL_MS = 20;
 const AGENT_CLOSE_TERM_GRACE_MS = 1_500;
 const AGENT_CLOSE_KILL_GRACE_MS = 1_000;
 const STARTUP_STDERR_MAX_CHARS = 8_192;
+const RUNTIME_STDERR_MAX_CHARS = 8_192;
 
 type LoadSessionOptions = {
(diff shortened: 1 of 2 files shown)


steipete:
Codex maintainer review: not landing this one as-is.

The ACP payload-message path is reasonable, but the stderr fallback is too broad: any generic `-32603 Internal error` can become the last 8 KiB of adapter stderr in the user-facing prompt error. That can surface unrelated logs, noisy stack traces, or sensitive provider/plugin output that was not otherwise part of the structured ACP error.

A safer version would either:

- only use structured ACP `error.data.message` / known structured fields, or
- use an allowlisted parser for specific known adapter patterns and scrub/truncate aggressively, with tests proving unrelated stderr is ignored.

#256 now landed the lower-risk first pass for text-mode remediation hints while keeping JSON/machine error payloads stable. This PR still has a useful idea, but it needs the narrower error-detail boundary before merge.

PR #265, opened 2026-04-21
fix(runtime): retain persistent ACP client across turns

### LINKED ISSUE #264: [bug] persistent ACP client not returned to pool after runTurn
## Version
- `openclaw` 2026.4.15 (npm global install, dist inspected under `~/.local/lib/node_modules/openclaw/dist/`)
- `openclaw/acpx` `src/runtime/engine/manager.ts@main`
- Reproduced with `claude-agent-acp` and `gemini --acp`; affects all ACP agents

## Summary
Persistent ACP sessions pay a full cold-start cost on every `sessions_send`. `acp.runtime.ttlMinutes` has no effect on adapter process longevity for the persistent path because the adapter is torn down after every turn, regardless of session mode.

Root cause is a lifecycle asymmetry between `ensureSession` and `runTurn`:

- `ensureSession` (around L293) puts the live client into `pendingPersistentClients` when `mode === "persistent"` and marks `keepClientOpen = true`.
- `runTurn` (around L324) takes the client out of the pool at turn start.
- `runTurn` `finally` (around L519) **unconditionally** `await client.close()`, with no symmetric "return to pool" branch for persistent sessions.

Net effect: the pool entry exists for exactly one turn, then the adapter dies.

## Reproduction

```
# ttlMinutes = 30 in acp.runtime config
spawn persistent label=verify-reuse-v1 agent=gemini-acp
# adapter child visible in ps, e.g. PID 104981
send message (turn 1)
# after turn completes:
#  without fix → adapter exits shortly, next send re-spawns
#  with fix    → PID 104981 retained in pool
send message (turn 2)
# agent_started_at in persisted session record:
#  without fix → new timestamp, new PID, full cold-start (~23s for Claude, 74-112s for Gemini)
#  with fix    → same timestamp, same PID, warm path
```

## Code references

`src/runtime/engine/manager.ts`:

```ts
// L293 — ensureSession puts client into pool when persistent
if (input.mode === "persistent") {
  const previousClient = this.pendingPersistentClients.get(record.acpxRecordId);
  this.pendingPersistentClients.set(record.acpxRecordId, client);
  keepClientOpen = true;
  await previousClient?.close().catch(() => {});
}

// L324 — runTurn takes the client out of the pool
let pendingClient = this.pendingPersistentClients.get(record.acpxRecordId);
if (pendingClient) {
  this.pendingPersistentClients.delete(record.acpxRecordId);
  // ...
}

// L519 — runTurn finally closes unconditionally (NO symmetric return-to-pool)
} finally {
  // ...
  await this.options.sessionStore.save(record).catch(() => {});
  await client.close().catch(() => {});   // ← persistent client lost here
  queue.close();
}
```

## Proposed fix

Two edits. Both preserve existing behavior for oneshot records (recordId contains `":oneshot:"`) and for sessions whose backend reports no reusable handle.

### Edit 1 — `runTurn` finally: conditionally return client to pool

```ts
} finally {
  // ...
  await this.options.sessionStore.save(record).catch(() => {});

  const isPersistentRecord = !record.acpxRecordId.includes(":oneshot:");
  let pooled = false;
  if (
    isPersistentRecord &&
    !record.closed &&
    client.hasReusableSession(record.acpSessionId)
  ) {
    const previousClient = this.pendingPersistentClients.get(record.acpxRecordId);
    this.pendingPersistentClients.set(record.acpxRecordId, client);
    pooled = true;
    if (previousClient && previousClient !== client) {
      await previousClient.close().catch(() => {});
    }
  }
  if (!pooled) await client.close().catch(() => {});
  queue.close();
}
```

### Edit 2 — `close()`: clean pool on non-discard path

`AcpxManager.evictIdleRuntimeHandles` (in openclaw core) calls `runtime.close({ handle, reason: "idle-evicted" })` **without** `discardPersistentState`. Without this edit the pool leaks CLI processes along the idle-evicted path.

```ts
async close(
  handle: AcpRuntimeHandle,
  options: { discardPersistentState?: boolean } = {},
): Promise<void> {
  const record = await this.requireRecord(handle.acpxRecordId ?? handle.sessionKey);
  await this.cancel(handle);
  if (options.discardPersistentState) {
    await this.closeBackendSession(record);
    record.acpx = { ...record.acpx, reset_on_next_ensure: true };
  } else {
    const pendingClient = this.pendingPersistentClients.get(record.acpxRecordId);
    if (pendingClient) {
      this.pendingPersistentClients.delete(record.acpxRecordId);
      await pendingClient.close().catch(() => {});
    }
  }
  record.closed = true;
  record.closedAt = isoNow();
  await this.options.sessionStore.save(record);
}
```

## Local validation

Applied equivalent edits to dist (`register.runtime-*.js`), restarted gateway, then ran two turns against a single persistent `gemini-acp` label:

| Field | Turn 1 | Turn 2 | Result |
|---|---|---|---|
| `pid` | 104981 | 104981 | reused ✓ |
| `agent_started_at` | `08:59:24.468Z` | `08:59:24.468Z` | stable ✓ |
| `last_used_at` | `08:59:51Z` | `09:00:59Z` | updated ✓ |
| turn latency | ~19s (first prompt; Gemini cold inference) | ~17s (warm) | Gemini CLI inference time dominates |

Process verified via `ps -p 104981`:

```
PID     ELAPSED  CMD
104981  45:49    node gemini -m gemini-3-flash-preview -y --acp
```

Retained across both turns; reaped on next unrelated `spawn` which triggered `evictIdleRuntimeHandles` (confirming Edit 2 is necessary — without it the pool would leak on the idle-evicted close path).

## Environment
- Linux 6.17.0-20-generic
- Gateway version: `2026.4.15`
- ACP agents configured: 6 (3 claude, 3 gemini) via `acp.allowedAgents`

Sway-Chan:
## Summary
Fixes #264.

Two edits to `src/runtime/engine/manager.ts`:

1. `runTurn` finally now returns the client to `pendingPersistentClients` when the record is persistent (recordId does not contain `":oneshot:"`) and the backend reports a reusable session, restoring lifecycle symmetry with `ensureSession`.
2. `close()` now cleans the pool on the non-`discardPersistentState` path, so `AcpxManager.evictIdleRuntimeHandles` (reason=`"idle-evicted"`) does not leak pooled CLI processes.

## Test plan
- [x] Manual functional verification via equivalent edits to the distributed `register.runtime-*.js`, then restarting the gateway: two turns against the same persistent `gemini-acp` label; `pid` and `agent_started_at` stable across turns; pooled process reaped on next unrelated spawn (see issue for details).
- [ ] CI / upstream test suite (for maintainer review).

## Related
- Linked issue describes the lifecycle asymmetry, reproduction, and validation data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Repo: https://github.com/openclaw/acpx
 
 ### Fixes
 
+- Runtime/persistent sessions: keep reusable persistent ACP clients warm across turns and close pooled clients during runtime close. (#265) Thanks @Sway-Chan.
 - Runtime/ACP: drain late post-success session updates before closing prompt turns so adapters that resolve `session/prompt` before final updates do not drop assistant output. (#251) Thanks @logofet85-ai.
 - Sessions/reset: close the live backend session when discarding persistent state so reset flows start a fresh ACP session instead of silently reopening the old one.
 - Agents/kiro: use `kiro-cli-chat acp` for the built-in Kiro adapter command to avoid orphan child processes. (#129) Thanks @vokako.
--- a/src/runtime/engine/manager.ts
+++ b/src/runtime/engine/manager.ts
@@ -542,7 +542,9 @@ export class AcpRuntimeManager {
             await previousClient.close().catch(() => {});
           }
         }
-        if (!pooled) {await client.close().catch(() => {});}
+        if (!pooled) {
+          await client.close().catch(() => {});
+        }
         queue.close();
       }
     })();
--- a/src/runtime/engine/manager.ts
+++ b/src/runtime/engine/manager.ts
@@ -528,7 +528,21 @@ export class AcpRuntimeManager {
         applyConversation(record, conversation);
         record.lastUsedAt = isoNow();
         await this.options.sessionStore.save(record).catch(() => {});
-        await client.close().catch(() => {});
+        const isPersistentRecord = !record.acpxRecordId.includes(":oneshot:");
+        let pooled = false;
+        if (
+          isPersistentRecord &&
+          !record.closed &&
+          client.hasReusableSession(record.acpSessionId)
+        ) {
+          const previousClient = this.pendingPersistentClients.get(record.acpxRecordId);
+          this.pendingPersistentClients.set(record.acpxRecordId, client);
+          pooled = true;
+          if (previousClient && previousClient !== client) {
+            await previousClient.close().catch(() => {});
+          }
+        }
+        if (!pooled) {await client.close().catch(() => {});}
         queue.close();
       }
     })();
@@ -640,6 +654,12 @@ export class AcpRuntimeManager {
         ...record.acpx,
         reset_on_next_ensure: true,
       };
+    } else {
+      const pendingClient = this.pendingPersistentClients.get(record.acpxRecordId);
+      if (pendingClient) {
+        this.pendingPersistentClients.delete(record.acpxRecordId);
+        await pendingClient.close().catch(() => {});
+      }
     }
     record.closed = true;
     record.closedAt = isoNow();
--- a/test/runtime-manager.test.ts
+++ b/test/runtime-manager.test.ts
@@ -312,6 +312,110 @@ test("AcpRuntimeManager streams runtime events and saves updated status", async 
   assert.equal(saved?.protocolVersion, 1);
 });
 
+test("AcpRuntimeManager keeps reusable persistent clients pooled across turns and closes them on runtime close", async () => {
+  const record = makeSessionRecord({
+    acpxRecordId: "pooled-persistent-session",
+    acpSessionId: "pooled-sid",
+    agentCommand: "gemini --acp",
+    cwd: "/workspace",
+  });
+  const store = new InMemorySessionStore([record]);
+  let factoryCalls = 0;
+  let closeCalls = 0;
+  let promptCalls = 0;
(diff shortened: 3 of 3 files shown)


enki:
Thanks for digging into this. We confirmed the lifecycle bug affects our `proving` use of `acpx/runtime`.

I ported the same fix into our active runtime branches, with two extra regression cases:

- persistent turns retain and reuse the same live ACP client
- `close()` racing an active persistent turn does not allow the turn finalizer to re-pool the client or overwrite the closed state

Updated branches/PRs:

- #262: runtime turn handles
- #263: runtime control session reuse

Both branch-specific runtime-manager suites pass, and the full `pnpm test` passed on rerun for both branches. The local integration branch has the same port as well.

So yes, this was a real bug and the core fix direction in this PR is correct. The only practical difference is that our current branches needed the fix applied to their refactored turn lifecycle path rather than cherry-picking this patch directly.

steipete:
Landed via maintainer fixup and rebase onto main.

- Gate: `pnpm run check` passed locally (format, typecheck, lint, build, viewer type/build, coverage: 530 tests)
- Targeted proof: persistent runtime manager regression covers two turns reusing one client and runtime close reaping the pooled client
- Land commit: https://github.com/openclaw/acpx/commit/9a540808bd039f770eb94e3ec504a0e5d2dca052
- Source fixup commit before rebase: https://github.com/Sway-Chan/acpx/commit/715141a7ba2032a552e7f3a9de414847d5651da0

Thanks @Sway-Chan!

PR #282, opened 2026-04-28
feat(runtime): expose advertised configOptions per-session via handle-aware getCapabilities

samithaj:
## Summary

Capture the `configOptions` array adapters return on `session/new` and `session/load` and surface those advertised ids through `getCapabilities({ handle })`. This lets embedders intersect their intended `session/set_config_option` pairs against the advertised set before issuing the RPC, rather than discovering unsupported keys via `-32603 Unknown config option` rejections.

The motivating case is `@agentclientprotocol/claude-agent-acp@0.31+`, which only advertises `mode | model | effort`. Embedders such as OpenClaw that minted `timeout`/`thinking`/`approval_policy` against a generic vocabulary were getting opaque `ACP_TURN_FAILED` turns because there was no per-session capability surface to filter against.

## What changes

- `SessionCreateResult` and `SessionLoadResult` now carry an optional `configOptions?: SessionConfigOption[]`, populated from the SDK's `NewSessionResponse.configOptions` and `LoadSessionResponse.configOptions`.
- `manager.ensureSession()` and `reconnect.connectAndLoadSession()` (both load and fallback-create branches) call `applyConfigOptionsToRecord(record, ...)` so the advertised list is persisted to `record.acpx.config_options`.
- `applyConfigOptionsToRecord` is promoted from a private function inside `manager.ts` to a small shared helper at `src/runtime/engine/session-config-options.ts` so both call sites use the same clone-and-write contract.
- `AcpxRuntime.getCapabilities()` becomes handle-aware:
  - `getCapabilities()` and `getCapabilities({})` (no handle) are still **synchronous** and return the same static `ACPX_CAPABILITIES`.
  - `getCapabilities({ handle })` returns a `Promise<AcpRuntimeCapabilities>` whose `configOptionKeys` is populated from `record.acpx.config_options.map(opt => opt.id)` (deduped). When the record has no `config_options` it falls back to base capabilities.

The public `AcpRuntime.getCapabilities` contract already declared the return as `Promise<AcpRuntimeCapabilities> | AcpRuntimeCapabilities` and `AcpRuntimeCapabilities.configOptionKeys?: string[]` was already part of the type — so this is fully additive. Patch bump (`0.6.1 → 0.6.2`).

## Why pair with `loadSessionWithOptions(..., { suppressReplayUpdates: true })`

`reconnect.connectAndLoadSession` deliberately suppresses replayed `config_option_update` notifications during load, so the `setConfigOption` response handler (which already calls `applyConfigOptionsToRecord`) cannot be relied on to seed the advertised list on resume. Capturing from `LoadSessionResponse.configOptions` directly is the only reliable seed.

## Repro that motivated this

Verified end-to-end against `npx -y @agentclientprotocol/claude-agent-acp@0.31` over stdio:

```
[repro] step 2: session/new
[repro] session/new result keys: [ 'sessionId', 'models', 'modes', 'configOptions' ]
[repro] step 3: set_config_option { configId: "timeout", value: "120" }
{ code: -32603, message: "Internal error", data: { details: "Unknown config option: timeout" } }
[repro] step 4 (sanity): set_config_option { configId: "mode", value: "default" }
  → succeeds
```

The harness exposes `configOptions` on `session/new` (here advertising `mode`, `model`, `effort`); without this PR the runtime drops that information on the floor.

## Test plan

- [x] `pnpm run lint && pnpm run typecheck && pnpm test` (590 tests, all passing)
- [x] New tests in `test/runtime.test.ts`:
  - `getCapabilities()` and `getCapabilities({})` stay synchronous and return base capabilities.
  - `getCapabilities({ handle })` returns advertised `configOptionKeys` from the record.
  - `getCapabilities({ handle })` returns base capabilities when the record has no `config_options`.
- [x] New tests in `test/runtime-manager.test.ts`:
  - `manager.ensureSession` populates `record.acpx.config_options` from `session/new`.
  - `manager.ensureSession` populates `record.acpx.config_options` from `session/load` on resume.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Repo: https://github.com/openclaw/acpx
 
 ### Changes
 
+- Runtime/embedding: capture per-session advertised `configOptions` from `session/new` and `session/load` and surface them via handle-aware `getCapabilities`, so embedders can reconcile advertised keys before issuing `session/set_config_option` against adapters with narrow advertised configOption sets (e.g. `claude-agent-acp` v0.31+ which only advertises `mode | model | effort`). `getCapabilities()` and `getCapabilities({})` keep their synchronous return; only `getCapabilities({ handle })` becomes async. Thanks @samithaj.
+
 ### Breaking
 
 ### Fixes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "acpx",
-  "version": "0.6.1",
+  "version": "0.6.2",
   "description": "Headless CLI client for the Agent Client Protocol (ACP) — talk to coding agents from the command line",
   "keywords": [
     "acp",
--- a/src/acp/client.ts
+++ b/src/acp/client.ts
@@ -18,6 +18,7 @@ import {
   type ReleaseTerminalResponse,
   type RequestPermissionRequest,
   type RequestPermissionResponse,
+  type SessionConfigOption,
   type SessionNotification,
   type SetSessionConfigOptionResponse,
   type TerminalOutputRequest,
@@ -116,11 +117,13 @@ export type SessionCreateResult = {
   sessionId: string;
   agentSessionId?: string;
   models?: SessionModelState;
+  configOptions?: SessionConfigOption[];
 };
 
 export type SessionLoadResult = {
   agentSessionId?: string;
   models?: SessionModelState;
+  configOptions?: SessionConfigOption[];
 };
 
 type AgentDisconnectReason = "process_exit" | "process_close" | "pipe_close" | "connection_close";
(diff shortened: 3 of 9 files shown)


steipete:
Landed the useful part in https://github.com/openclaw/acpx/commit/43013ead1ab84bceaa1fe4552743c3fb108a5368.

What changed:
- `AcpClient` now returns `configOptions` from `session/new` and `session/load`.
- Session records persist those advertised options.
- Embedded runtime `getCapabilities({ handle })` exposes stable `configOptionKeys` for the resolved handle.
- Avoided the package version bump and used the existing encoded-handle resolution path.

Validation:
- Checked against the ACP config options spec: initial `configOptions` may be returned during session setup and clients should use them.
- Reproduced missing persisted options locally with the mock agent.
- Live CLI/runtime probe returned `configOptionKeys: ["mode", "model", "reasoning_effort"]`.

Thanks @samithaj, and credited in the changelog.

PR #283, opened 2026-04-29
fix: send session/close before agent termination

codefromthecrypt:
Migrate from nes/close to session/close. Add sendSessionClose option to AcpClient.close() so callers that end a session can request session/close before the agent is killed, while callers that detach (e.g. sessions new) skip it. Add a SIGTERM handler to the queue owner so its finally block runs when sessions close sends the signal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Repo: https://github.com/openclaw/acpx
 ### Fixes
 
 - CLI/models: fail clearly when `--model` targets a non-Claude ACP agent that does not advertise ACP model support, and reject model ids outside an adapter's advertised `availableModels` instead of silently falling back to the adapter default.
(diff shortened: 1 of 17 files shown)


steipete:
Landed a narrower version in https://github.com/openclaw/acpx/commit/43013ead1ab84bceaa1fe4552743c3fb108a5368.

I did not merge this branch as-is because the final shape made plain `AcpClient.close()` send ACP `session/close`. Per the ACP session lifecycle, `session/close` frees active session resources, while `session/load` / `session/resume` are the continuity paths. Sending close from generic detach/reconnect cleanup could invalidate persistent sessions.

What landed instead:
- Added a queue-owner `close_session` IPC control.
- `sessions close` asks the active owner to send ACP `session/close` before terminating the adapter process.
- Plain `AcpClient.close()` remains non-destructive for detach/reconnect paths.

Validation:
- Reproduced the missing close notification with a live mock agent marker.
- Live `sessions close` wrote the backend session id through mock `session/close` before shutdown.

Thanks @codefromthecrypt, and credited in the changelog.

codefromthecrypt:
thanks tons @steipete!

codefromthecrypt:
@steipete sorry to nag, but is the next version intentionally tagged but not a release?

PR #288, opened 2026-05-01
feat: forward detailCode on AcpRuntimeTurnResultError

kunchenguid:
## Summary

Adds an optional `detailCode` field to `AcpRuntimeTurnResultError`, forwarded from the existing `NormalizedOutputError`. acpx already classifies errors internally (`AUTH_REQUIRED`, `AGENT_DISCONNECTED`, `AGENT_STARTUP_FAILED`, `SESSION_RESUME_REQUIRED`, ...), but `src/runtime/engine/manager.ts:716-726` was dropping the classification before the result reached consumers.

## Why

Embedding `acpx/runtime` in a long-running coding-agent loop. The harness needs to distinguish "permanent failure, give up immediately" from "transient error, retry with backoff" without substring-matching `error.message`. `retryable` is one bit and collapses cases that warrant different harness behavior - `AUTH_REQUIRED` is terminal, `AGENT_STARTUP_FAILED` often means "tell the user to install the adapter," both are non-retryable but the right response differs.

Reference consumer: https://github.com/kunchenguid/gnhf

## Tests

- updated the existing "fails persistent turns clearly" assertion - `SESSION_RESUME_REQUIRED` was already being computed internally and is now visible at the result boundary, which is the change in shape
- new dedicated test "forwards normalized detailCode on turn failures" using `AcpxOperationalError` with a known `detailCode`

## Open questions

**Legacy event variant.** There's a parallel `type: "error"` event in `src/runtime/public/contract.ts:115-120` emitted by `runTurn(...)` for back-compat consumers. For symmetry, `detailCode` could appear there too. Left it out of this PR to keep the surface area small. Happy to follow up if maintainers want symmetry, or to leave the legacy event frozen on purpose.

**`string` vs typed union.** `detailCode: string` matches the internal type. A typed union (`"AUTH_REQUIRED" | "AGENT_DISCONNECTED" | ...`) at the public boundary would be friendlier to consumers but locks the values in. Went with `string` (easier to evolve), happy to be talked out of it.

### DIFF
--- a/src/runtime/engine/manager.ts
+++ b/src/runtime/engine/manager.ts
@@ -722,6 +722,7 @@ export class AcpRuntimeManager {
             message: normalized.message,
             ...(normalized.code ? { code: normalized.code } : {}),
             ...(normalized.retryable !== undefined ? { retryable: normalized.retryable } : {}),
+            ...(normalized.detailCode ? { detailCode: normalized.detailCode } : {}),
           },
         });
       } finally {
--- a/src/runtime/public/contract.ts
+++ b/src/runtime/public/contract.ts
@@ -123,6 +123,7 @@ export type AcpRuntimeTurnResultError = {
   message: string;
   code?: string;
   retryable?: boolean;
+  detailCode?: string;
 };
 
 export type AcpRuntimeTurnResult =
--- a/test/runtime-manager.test.ts
+++ b/test/runtime-manager.test.ts
@@ -1,6 +1,7 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 import type { SetSessionConfigOptionResponse } from "@agentclientprotocol/sdk";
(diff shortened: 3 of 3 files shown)


steipete:
Landed the useful part in https://github.com/openclaw/acpx/commit/43013ead1ab84bceaa1fe4552743c3fb108a5368.

What changed:
- `AcpRuntimeTurnResultError` now preserves normalized `detailCode`.
- The `runTurn(...)` compatibility error event forwards the same detail code.
- Added regression coverage for the failed-turn result and legacy event path.

Validation:
- Reproduced the loss locally with the runtime failure path.
- Checked the ACP/runtime error flow against current source.
- Live runtime probe returned `detailCode: SESSION_RESUME_REQUIRED` through `startTurn(...).result`.

Thanks @kunchenguid, and credited in the changelog.

PR #314, opened 2026-05-12
fix: checkpoint live session updates during prompt turns

### LINKED ISSUE #304: ACP session history only records prompts while Codex is running
## Summary

For an actively running prompt turn, `acpx <agent> sessions history <name>` only shows the user prompt entry. Assistant text and tool-call entries do not appear in `sessions history` until the turn finishes — even though the underlying agent (codex in this case) has already produced them and they are visible in the Codex rollout log on disk.

This is closely related to (but distinct from) #157, which fixed the prompt-checkpointing race so `0/0 No history` no longer appears at the start. That fix is working — the prompt is checkpointed up-front. The remaining gap is that subsequent assistant/tool events are not checkpointed as they stream in.

## Repro

Basic foreground prompt:

1. `acpx codex sessions new --name foo`
2. Send a long-running prompt that uses tools, e.g.:
   ```bash
   acpx codex -s foo "list 10 files in this repo, then read each one and summarize"
   ```
3. While the prompt is still running, in another terminal:
   ```bash
   acpx codex sessions history foo --limit 20
   ```

Queued stdin prompt, matching the programmatic orchestration case:

```bash
acpx codex sessions ensure --name foo

acpx codex -s foo --no-wait --file - <<'PROMPT'
Do a long-running coding task that reads files, edits code, and runs tests.
PROMPT

acpx codex -s foo status
acpx codex sessions history foo --limit 20
acpx codex sessions read foo --tail 20
```

## Expected

History shows the user prompt **and** the assistant/tool entries that have already been produced by the agent during the turn, similar to what's already in the Codex rollout JSONL.

## Actual

Only the user prompt is shown:

```text
session: <id> (1/1 shown)
2026-05-08T...   user   <prompt>
```

The session `agentSessionId` field also remains `-` while the turn is in flight, which makes it harder for an external orchestrator to reliably tail progress via the persisted record.

In the queued-stdin run, this was not an agent-startup issue:

- `acpx codex -s <name> status` showed the queue owner as running.
- `ps` showed the queue owner had launched `npm exec codex-acp` and the `codex-acp` binary.
- `lsof -p <codex-acp-pid>` showed `codex-acp` actively writing a Codex rollout log, e.g. `~/.codex/sessions/YYYY/MM/DD/rollout-....jsonl`.
- Tailing that rollout JSONL showed assistant messages, tool calls, tool outputs, test output, and file edits.
- At the same time, `acpx codex sessions history ...` and `sessions read ...` still showed only the initial user prompt.

So the underlying agent is active and producing events; the missing piece is that those live streamed assistant/tool events are not reflected through the acpx session history/read APIs during the turn.

## Why it matters

When using acpx as a programmatic orchestration layer (e.g. driving multiple parallel codex sessions from another agent), `sessions history` is the natural place to poll for live progress without attaching to the foreground stdout stream. Today the rollout log on disk has the data but `sessions history` does not, so consumers have to fall back to parsing the rollout log directly.

The queued stdin shape (`--no-wait --file -`) is especially important because it is the natural API shape for detached/background orchestration.

## Environment

- acpx 0.7.0 (latest)
- agent: `npx @zed-industries/codex-acp@^0.12.0`
- macOS (Darwin 25.3.0)

## Note

After #157's fix (`013f2fd`, 2026-04-25), the prompt is now checkpointed at submit time. This issue is asking for the same checkpointing pattern to be extended to assistant/tool events as they stream through the ACP runtime, so the persisted session record is a complete live picture of the turn rather than only its bookends.

AndroidPoet:
Fixes #304.

## Summary

This makes the persisted session record track live prompt progress instead of only the prompt submission and final turn state.

The change adds a small debounced checkpoint helper and wires it into both prompt execution paths:

- foreground/session prompt runs through `src/cli/session/runtime.ts`
- detached queue-owner/background runs through `src/runtime/engine/manager.ts`

When ACP `session/update` events arrive, the in-memory conversation is still updated first, but now a live checkpoint is scheduled after that mutation. The checkpoint also flushes pending ACP wire messages before saving, so the event stream and session record move forward together.

There is also an immediate checkpoint after reconnect/load resolution. That makes fields like `agentSessionId` visible before the prompt finishes, instead of waiting for the turn close path.

## Why this shape

I avoided reading Codex rollout JSONL files. That would fix the reported Codex case, but it would make `acpx` depend on one adapter's private logs. This stays inside the ACP runtime and applies to any adapter that emits `session/update` notifications.

The live saves are debounced at 500ms so token-by-token streams do not write the session JSON file for every chunk. Boundary saves still flush immediately where it matters: after session resolution and on close/finalization.

## Tests

- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run test`

Added integration coverage for:

- `sessions read` seeing assistant output while a foreground prompt is still sleeping
- `--no-wait --file -` stdin prompts checkpointing assistant output before the detached turn finishes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Repo: https://github.com/openclaw/acpx
 ### Fixes
 
 - Runtime/embedding: preserve structured ACP `tool_call_update` details on public runtime events, including content, output, locations, kind, and raw payload fields, so embedders can display live tool progress. (#306) Thanks @joeia26.
+- CLI/sessions: checkpoint live assistant and tool updates while prompt turns are still running, so `sessions read` and `sessions history` can show in-flight progress instead of only the submitted prompt. (#314) Thanks @AndroidPoet.
 
 ## 2026.5.5 (v0.7.0)
 
--- a/src/cli/session/runtime.ts
+++ b/src/cli/session/runtime.ts
@@ -29,6 +29,7 @@ import {
   trimConversationForRuntime,
 } from "../../session/conversation-model.js";
 import { SessionEventWriter } from "../../session/events.js";
+import { LiveSessionCheckpoint } from "../../session/live-checkpoint.js";
 import { setCurrentModelId, setDesiredModelId } from "../../session/mode-preference.js";
 import {
   absolutePath,
@@ -408,6 +409,22 @@ async function runSessionPrompt(options: RunSessionPromptOptions): Promise<Sessi
       await eventWriter.appendMessages(batch, { checkpoint });
     });
   };
+  const liveCheckpoint = new LiveSessionCheckpoint({
+    save: async () => {
+      await flushPendingMessages(false);
+      record.lastUsedAt = isoNow();
+      applyConversation(record, conversation);
+      record.acpx = acpxState;
+      await eventWriter.checkpoint();
+    },
+    onError: (error) => {
+      if (options.verbose) {
+        process.stderr.write(
+          "[acpx] live session checkpoint failed: " + formatErrorMessage(error) + "\n",
+        );
+      }
+    },
+  });
 
   const ownClient = options.client == null;
   const client =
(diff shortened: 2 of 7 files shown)


steipete:
Landed via rebase onto main.

- Gate: pnpm run check
- Review: codex-review clean, no accepted/actionable findings
- Land commit: 27d6a221c0065f67bd2f8eba729f3170636ce533
- Merge commit: 0f7d65d6123388fc011e658d08bc6a2a9a14b94c

Thanks @AndroidPoet!

PR #326, opened 2026-05-19
feat(json-strict): default --prompt-retries to 1 when strict mode is set

coletebou:
## Summary

`--json-strict` signals "I'm a programmatic caller, don't be flaky on transient stream cuts." But `--prompt-retries` defaults to 0 even in strict mode, so a single transient failure crashes the caller. This PR makes the retry default depend on `jsonStrict`:

- `--json-strict` AND `--prompt-retries` not provided → default 1
- `--json-strict` AND `--prompt-retries 0` (explicit) → 0 (honored)
- `--json-strict` AND `--prompt-retries N` → N (unchanged)
- non-strict mode → unchanged (default `undefined` → downstream `?? 0`)

Detection uses `typeof opts.promptRetries === "number"` since commander has no `.default()` on this option, so `undefined` cleanly indicates "not provided."

## Files

- `src/cli/flags.ts` — one ternary in `resolveGlobalFlags`; help text updated to `(default: 0, or 1 with --json-strict)`
- `test/cli-flags.test.ts` — 4 new cases covering the matrix
- `test/integration.test.ts` — one pre-existing test renamed and updated to pass `--prompt-retries 0` explicitly (it asserted "no retry" under the old default; under the new default the same assertion holds with the explicit flag)

## Why

A previous run of `clawpatch` with acpx as the provider produced 4 distinct unparseable-output failures across 100 features, several of which would have cleared on a single retry. Strict mode is exactly where opinionated defaults make sense.

## Validation

- `pnpm format:check` — clean
- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm build` — clean
- `pnpm viewer:build` — clean
- `pnpm test:coverage` — 675/675 pass (was 674/675 with one regression caught and fixed)

## Notes

- Smallest acpx PR in this stack — single resolver branch.
- Companion to the `--require-stop-reason` and terminal-envelope PRs.
- Explicit user value always wins (including `0`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Repo: https://github.com/openclaw/acpx
 
 ### Changes
 
+- CLI: when `--json-strict` is set and `--prompt-retries` is not explicitly provided, the default for `--prompt-retries` is now `1` (was `0`). Strict mode signals a programmatic caller, and surviving one transient stream cut by default is sane. Explicit `--prompt-retries 0` is still honored — operators who want to fail fast on the first transient error can. Non-strict mode is unchanged (default remains `0`).
+
 ### Breaking
 
 ### Fixes
--- a/src/cli/flags.ts
+++ b/src/cli/flags.ts
@@ -289,7 +289,7 @@ export function addGlobalFlags(command: Command): Command {
     )
     .option(
       "--prompt-retries <count>",
-      "Retry failed prompt turns on transient errors (default: 0)",
+      "Retry failed prompt turns on transient errors (default: 0, or 1 with --json-strict)",
       parsePromptRetries,
     )
     .option(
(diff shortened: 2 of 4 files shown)


vincentkoc:
Thanks for the PR. I’m going to close this one because the behavior change is too implicit for the CLI contract.\n\n`--json-strict` is an output-channel guarantee today. Making it silently change prompt retry behavior from 0 to 1 mixes output formatting with execution policy, which can surprise callers that intentionally want fail-fast behavior unless they remember to add `--prompt-retries 0`. My preference is to keep retries explicit through `--prompt-retries`.\n\nIf we decide strict JSON should imply a reliability profile later, that should be a deliberate product decision with docs spelling out the policy and tests covering the real CLI/queue behavior. Thanks for exploring it.

PR #327, opened 2026-05-19
feat(json-strict): always emit terminal prompt_result envelope

coletebou:
## Summary

Programmatic callers running with `--json-strict` need a single, structured signal that a prompt has finished. Today, when a prompt ends without producing any `agent_message_chunk` (e.g. `stopReason: cancelled` after a permission denial), consumers see only `[available_commands_update, usage_update]` envelopes and have to infer "no model output" from absence — which looks identical to a real envelope-shape regression.

This PR guarantees that `--json-strict` always emits exactly one terminal `acpx/prompt_result` envelope per prompt as the last line of stdout, regardless of whether message chunks were produced.

## Envelope shape

```json
{"jsonrpc":"2.0","method":"acpx/prompt_result","params":{
  "stopReason":"end_turn",
  "sessionId":"...",
  "finalText":"concatenated agent_message_chunk text, or null when stopReason!=end_turn",
  "hadMessageChunk":true,
  "messageChunkCount":7,
  "thoughtChunkCount":3,
  "elapsedMs":12345
}}
```

The existing `toPromptResult` in `src/cli/session/runtime.ts:133` already produces the shape internally; this PR just emits it as a discrete JSON-RPC notification when `jsonStrict` is set.

## Why

Built for programmatic consumers like `clawpatch` that need to distinguish:
- "model returned end_turn but malformed JSON" → real failure
- "model was cancelled / refused / hit max_tokens" → different failure class

Without a terminal envelope, the consumer has to scan the entire stream and infer state. With one, it reads the last line and branches on `stopReason`.

## Files

- `src/types.ts` — `OutputFormatter.beginPrompt?` and `emitPromptResultEnvelope?` optional methods, plus `PromptResultEnvelopeInput` type
- `src/cli/output/json-formatter.ts` — `JsonOutputFormatter` gains a `jsonStrict` flag, tracks chunks + start time, emits the envelope
- `src/cli/output/output.ts` — threads `jsonStrict` through `createOutputFormatter`
- `src/cli/command-handlers.ts` — `handlePrompt` + `handleExec` pass `outputPolicy.jsonStrict`
- `src/cli/session/runtime.ts` — `runSessionPrompt` + `runOnce` call `output.beginPrompt?.()` before the prompt loop and `output.emitPromptResultEnvelope?.()` on success / error / cancel paths
- 3 new tests in `test/output.test.ts`: end_turn envelope, cancelled envelope (null finalText), non-strict negative

## Validation

- `pnpm format:check` — clean
- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm build` — clean
- `pnpm viewer:build` — clean
- `pnpm test:coverage` — full suite green (output.test.js: 18/18 with 3 new)

## Notes

- Non-strict mode emits nothing new (optional-chained calls are no-ops on text/quiet/queue/discard formatters).
- Non-`OutputFormatter` consumers see byte-identical output.
- Pairs cleanly with `--require-stop-reason` (separate PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Repo: https://github.com/openclaw/acpx
 
 ### Changes
 
+- CLI/output: `--json-strict` now emits a terminal `acpx/prompt_result` JSON-RPC envelope as the last line of stdout per prompt, summarizing `stopReason`, `sessionId`, accumulated `finalText` (only when `stopReason === "end_turn"`), `hadMessageChunk`, `messageChunkCount`, `thoughtChunkCount`, and `elapsedMs`. Programmatic callers should branch on this envelope's `stopReason` rather than inferring completion from chunk-stream tail shape. Cancelled and errored prompts now emit the envelope too, so consumers can distinguish "no chunks because failure" from "no chunks because parser ran off the end of the stream". Non-strict modes (`--format text`, `--format json` without `--json-strict`, `--format quiet`) are unchanged.
+
 ### Breaking
 
 ### Fixes
--- a/src/cli/command-handlers.ts
+++ b/src/cli/command-handlers.ts
@@ -293,6 +293,7 @@ export async function handlePrompt(
       sessionId: record.acpxRecordId,
     },
     suppressReads: outputPolicy.suppressReads,
+    jsonStrict: outputPolicy.jsonStrict,
   });
 
   await printPromptSessionBanner(record, agent.cwd, outputPolicy.format, outputPolicy.jsonStrict);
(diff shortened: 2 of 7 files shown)


steipete:
Closing per maintainer decision.

The underlying need is reasonable: programmatic callers need a deterministic way to know that a prompt is finished and why it finished. But this PR changes the documented `--format json --json-strict` contract from raw ACP JSON-RPC NDJSON into ACP plus an acpx-specific synthetic terminal event.

Main concerns:
- successful prompts already include the ACP JSON-RPC result with `stopReason`
- runtime failures already emit JSON-RPC error responses
- adding `acpx/prompt_result` creates a second terminal-ish signal in the stream
- `finalText` reconstruction is partial and would become a product contract
- the promised last-line behavior needs integration proof across exec, prompt, queue, cancel, permission denial, retries, and runtime errors

If there is a concrete failing command where today's stream lacks a reliable `stopReason` or error signal, we should fix that narrower bug or design an explicit acpx event-stream contract separately.

Thanks for the thoughtful PR.

PR #328, opened 2026-05-19
feat(cli): --require-stop-reason for typed exit codes

coletebou:
## Summary

Programmatic callers want a one-syscall signal — exit code — for "prompt finished cleanly" vs "prompt was rejected/cancelled/truncated". Today this requires parsing message text.

This PR adds `--require-stop-reason <reasons>` (CSV, repeatable). When set, acpx exits non-zero if the prompt's terminal `stopReason` isn't in the allowed set:

| stopReason | exit code |
|------------|----------:|
| (matches required) | 0 |
| `cancelled` | 7 |
| `refusal` | 8 |
| `max_tokens` / `max_turn_requests` | 9 |
| `timeout` (agent-reported) | 10 |
| (other) | 11 |

`PERMISSION_DENIED` (5) and `TIMEOUT` (3) infrastructure failures continue to win — the new logic only sets `process.exitCode` when it's not already non-zero.

## Files

- `src/types.ts` — extended `EXIT_CODES` with new entries (7-11); added `KNOWN_STOP_REASONS` allowlist + `exitCodeForStopReason()`
- `src/cli/flags.ts` — new `parseRequireStopReason` (CSV, repeatable, lowercased), `requireStopReason?: ReadonlySet<string>` on `GlobalFlags`
- `src/cli/command-handlers.ts` — `applyStopReasonExitCode` invoked after `applyPermissionExitCode` in both `handlePrompt` and `handleExec`
- `test/cli-stop-reason.test.ts` — 22 cases (mapping, parser CSV/repeat/case/empty/unknown, helper precedence)

## Why

`acpx` callers like `clawpatch` need to distinguish stopReason classes for retry logic and error reporting. With this flag, a script can branch on exit code instead of grepping output:

```bash
acpx ... --json-strict --require-stop-reason end_turn ...
case $? in
  0)  echo "ok" ;;
  7)  echo "cancelled — abort" ;;
  8)  echo "refused — review prompt" ;;
  9)  echo "truncated — increase budget" ;;
  *)  echo "infrastructure error" ;;
esac
```

## Validation

- `pnpm format:check` — clean
- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm build` — clean
- `pnpm viewer:build` — clean
- `pnpm test:coverage` — full suite green (cli-stop-reason.test.js: 22/22)

## Notes

- Fully additive — flag defaults off; existing behavior unchanged.
- Best paired with the terminal-envelope PR (which guarantees `stopReason` is reliably available in `--json-strict` mode).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Repo: https://github.com/openclaw/acpx
 
 ### Changes
 
+- CLI: add `--require-stop-reason <reason>` global flag so programmatic callers can gate prompt success on the terminal `stopReason` instead of inferring it from message chunks. The flag is repeatable and accepts comma-separated values; mismatches exit with a typed code (`STOP_REASON_CANCELLED=7`, `STOP_REASON_REFUSAL=8`, `STOP_REASON_LIMIT=9`, `STOP_REASON_TIMEOUT=10`, `STOP_REASON_OTHER=11`). Existing exit codes (0/1/2/3/4/5/130) and the `process.exitCode` set by permission denial are unchanged when the flag is omitted, and they win over a stopReason mismatch when set.
+
 ### Breaking
 
 ### Fixes
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/home/ubuntu/code/acpx/node_modules
--- a/src/cli/command-handlers.ts
+++ b/src/cli/command-handlers.ts
@@ -14,7 +14,7 @@ import {
   findSession,
   findSessionByDirectoryWalk,
 } from "../session/persistence.js";
-import { EXIT_CODES } from "../types.js";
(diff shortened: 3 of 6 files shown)


vincentkoc:
Thanks for putting this together. I’m going to close it rather than ask for incremental fixes here.\n\nThe useful idea is clear, but this PR is not mergeable as-is: it commits a `node_modules` symlink pointing at `/home/ubuntu/code/acpx/node_modules`, conflicts with current `main`, and adds a new CLI exit-code/product contract without the docs/CLI coverage and end-to-end proof we would need for a user-visible global flag.\n\nIf we revisit typed stop-reason gating later, I’d want it proposed as a smaller product-surface change first: no generated/local files, docs updated with the exact exit-code contract, integration coverage through the real CLI, and a fresh branch from current `main`. Thanks again for the contribution.

PR #424, opened 2026-07-04
fix(queue): don't orphan the agent bridge when the queue owner is terminated

superWorldSavior:
## Problem

When the queue owner process is terminated — SIGTERM from `terminateProcess`
in `lease-store`, or Ctrl-C/SIGINT — the `codex-acp` (or any) bridge adapter it
spawned can be left as an **orphaned child process**, accumulating across runs
until the agent app-server chokes and new sessions hang at `session/new`.

43013ea (which correctly preserves ACP session details and closes active owners)
addresses the **ACP session** lifecycle; the **OS-process** lifecycle of the
bridge is a separate issue, with three parts:

1. **No signal handler on the queue owner.** A SIGTERM killed the Node process
   immediately, bypassing the `finally` block, so `closeQueueOwnerRuntime()`
   (which kills the bridge via `sharedClient.close()`) never ran.
2. **Wrong shutdown order.** `owner.close()` (which `server.close()`s and waits
   for connected IPC client sockets to drain) ran *before* `sharedClient.close()`.
   A client parked in `waitForCompletion` kept its socket open, blocking the
   drain past the external SIGKILL grace — so the owner was SIGKILLed before it
   could kill the bridge.
3. **SIGTERM grace too short.** `terminateProcess` allowed 1.5 s, but a graceful
   `AcpClient.close()` needs ~2.6 s worst case (stdin-close 100 ms + SIGTERM
   1.5 s + SIGKILL 1 s).

## Fix

- Wrap the queue-owner main loop in `withInterrupt(...)`, installing
  SIGTERM/SIGINT/SIGHUP handlers that run graceful shutdown *before* the process
  exits.
- Share a single idempotent `shutdownPromise` between the signal path and the
  `finally` block, so `runSessionQueueOwner` never returns before
  `releaseQueueOwnerLease` completes.
- Reorder shutdown to **kill the bridge first**, then close the owner —
  invariant: *the bridge must never outlive the owner*. `SessionQueueOwner.close()`
  now also destroys tracked client sockets so `server.close()` resolves promptly
  even with a client still connected (clients already treat an unexpected
  disconnect as retryable).
- Raise the external SIGTERM grace to 4 s to cover the graceful-close budget with
  headroom.

**This does not touch `src/acp/*` or the semantics of `AcpClient.close()`** — it
is purely queue-owner / process-lifecycle. Plain `AcpClient.close()` stays
non-destructive and never sends `session/close`. (Deliberately scoped to avoid
the concern raised on #283.)

## Shutdown window

During bridge teardown the IPC server may briefly still accept a client against a
shutting-down session. Existing/late clients treat the disconnect as retryable;
this path is covered by a test. I can mark the owner closed earlier if you'd
prefer to reject immediately.

## Tests

`test/queue-owner-lifecycle.test.ts` (+ `test/queue-lease-store.test.ts`):
- exits 0 / releases lease on SIGTERM and SIGINT
- releases lease when `doShutdown` races between signal and finally
- kills the agent bridge when SIGTERMed mid-prompt
- **kills the bridge and exits gracefully when a client socket stays open during
  SIGTERM** (the exact orphan repro)

All green off `main`.

### DIFF
--- a/src/cli/queue/ipc-server.ts
+++ b/src/cli/queue/ipc-server.ts
@@ -122,6 +122,7 @@ export class SessionQueueOwner {
   private readonly onQueueDepthChanged?: (queueDepth: number) => void;
   private readonly pending: QueueTask[] = [];
   private readonly waiters: Array<(task: QueueTask | undefined) => void> = [];
+  private readonly activeSockets = new Set<net.Socket>();
   private closed = false;
 
   private constructor(
@@ -194,6 +195,18 @@ export class SessionQueueOwner {
       task.close();
     }
     this.emitQueueDepth();
+
+    // Destroy tracked sockets so server.close() can resolve promptly even
+    // when a client is still connected (e.g. in waitForCompletion).  Without
+    // this, server.close() waits for all existing connections to drain;
+    // a connected client would block the drain past the external SIGKILL grace,
+    // letting terminateProcess() kill the owner before the bridge was killed.
+    // Clients already handle unexpected disconnects gracefully (the error is
+    // caught on the client side and converted to a retryable error).
+    for (const socket of this.activeSockets) {
+      socket.destroy();
+    }
+    this.activeSockets.clear();
 
     await new Promise<void>((resolve) => {
       this.server.close(() => resolve());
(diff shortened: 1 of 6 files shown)


steipete:
Landed the repaired shutdown fix in https://github.com/openclaw/acpx/pull/430 as commit https://github.com/openclaw/acpx/commit/87327c6e4ac28243b94945afb922c6cbb0dd0f98.

The replacement was necessary because GitHub rejected authenticated maintainer pushes to this fork branch even though maintainer edits were enabled. The original contributor head remains untouched. Thank you @superWorldSavior for finding the bridge-orphaning shutdown defect and contributing the original fix; the landed commit preserves your co-author credit and the changelog thanks you.

PR #425, opened 2026-07-04
fix(output): surface a structured stderr line on error in quiet mode

superWorldSavior:
> Sibling to #424 (queue-owner orphan fix). Independent — touches only `output.ts` / `ipc.ts`, no file overlap with #424; either can land first.

## Problem

`QuietOutputFormatter.onError` is a no-op, so in `--format quiet` a failing
prompt is **silently swallowed**: a programmatic caller sees a non-zero exit with
empty stdout *and* empty stderr — indistinguishable from a bridge crash or a
provider blip. There is no signal at all that the prompt failed or why.

## Change

Emit exactly one structured line on **stderr** when a prompt fails in quiet mode:

```
[acpx] error: <code> [<detailCode>] <message collapsed to one line>
```

- **stdout / NDJSON is untouched.** This is stderr-only — the `--format json` /
  `--json-strict` contract is not modified and no synthetic event is added to the
  JSON-RPC stream. (Deliberately different from #327, which added a terminal
  envelope to the stdout stream.)
- The message is collapsed to a single line (`/\r\n?|\n/g`) so it stays
  grep/parse-friendly.
- On the queue IPC path, the error is marked `outputAlreadyEmitted` once the
  formatter has written it, so the top-level handler doesn't print a second raw
  line (no double emission).

## Precedent

Follows the spirit of #274 (quiet usage metadata): quiet carries structured
diagnostics on the side channel without polluting the data stream.

## On the contract change

I'm treating quiet as "no stdout data unless there is success output", not "no
diagnostics ever". This does change quiet's behavior from "fully silent" to
"silent on stdout, one diagnostic line on stderr for errors". If you'd rather
keep quiet strictly silent by default, I'm happy to:
- put it behind an opt-in flag (e.g. `--error-format=line`), or
- pursue dedicated exit codes (rate_limit / quota / internal / timeout) instead.

Flagging explicitly so the contract decision stays yours.

## Tests

`test/output.test.ts` (+ `test/queue-ipc-errors.test.ts`): structured line shape,
`\r\n` / bare `\r` / `\n` normalization, and no double-emission on the quiet IPC
path. All green off `main`.

### DIFF
--- a/src/cli/output/output.ts
+++ b/src/cli/output/output.ts
@@ -1146,7 +1146,7 @@ class QuietOutputFormatter implements OutputFormatter {
     }
   }
 
-  onError(_params: {
+  onError(params: {
     code: OutputErrorCode;
     detailCode?: string;
     origin?: OutputErrorOrigin;
@@ -1155,7 +1155,8 @@ class QuietOutputFormatter implements OutputFormatter {
     acp?: OutputErrorAcpPayload;
     timestamp?: string;
   }): void {
-    // no-op in quiet mode
+    const qualifier = params.detailCode ? `${params.code} ${params.detailCode}` : params.code;
+    this.stderr.write(`[acpx] error: ${qualifier} ${params.message}\n`);
   }
 
   onPermissionEscalation(_event: PermissionEscalationEvent): void {
--- a/src/cli/output/output.ts
+++ b/src/cli/output/output.ts
@@ -1156,7 +1156,7 @@ class QuietOutputFormatter implements OutputFormatter {
     timestamp?: string;
   }): void {
     const qualifier = params.detailCode ? `${params.code} ${params.detailCode}` : params.code;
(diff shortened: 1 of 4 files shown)


steipete:
Landed via https://github.com/openclaw/acpx/pull/431 in [`ffd8549`](https://github.com/openclaw/acpx/commit/ffd8549355adbb28bfc36ffd1087643d2df29fdb), including the regression coverage and maintainer changelog credit.

The contributor fork advertised maintainer edits, but GitHub rejected authenticated maintainer pushes to the branch. I therefore preserved the original head unchanged and recreated the proof-complete repair upstream with the original commit author retained as a co-author.

Thank you @superWorldSavior for the report and contribution! For future contributions, ensuring “Allow edits by maintainers” is enabled and accepted by the fork helps us apply required fixups directly.

PR #432, opened 2026-07-05
fix(client): reap the bridge process tree on close

superWorldSavior:
## Problem

When acpx terminates an agent bridge, `killAgentIfRunning` (`src/acp/client.ts`) called `child.kill(signal)` — signalling **only the bridge PID**. The bridge (codex-acp etc.) spawns its own children: the model process and MCP servers (e.g. `serena`). Those descendants were never signalled, so they survived as orphans and accumulated until the agent app-server choked and new sessions hung at `session/new`.

## Fix

On **POSIX**, spawn the bridge in its own process group (`detached`, `pgid == pid`) and signal the **whole group** on close (`process.kill(-pid, signal)`), gated on an `agentGroupLeader` flag so acpx never signals its own group. Windows and any non-detached path keep the previous single-PID `child.kill`.

A **final best-effort group sweep** also runs after the bridge PID itself has gone — e.g. a well-behaved bridge that exits during the stdin-close grace — so descendants that outlived the bridge are still reaped. Signal failures other than `ESRCH` are traced (verbose) rather than silently swallowed.

## Scope & limitations (explicit)

- **Process group, not full tree.** Reaps descendants that stay in the bridge's group. A descendant that re-detaches into its **own** session (`setsid`) leaves the group and is out of scope.
- **POSIX-only.** On Windows `agentGroupLeader` is false and behaviour is unchanged (no group reaping). A Job-Object path can follow.
- **Cooperative close only.** This couples descendants to `AcpClient.close()`. It does **not** cover abrupt death of acpx itself (`SIGKILL`/OOM) — there is no `PR_SET_PDEATHSIG` here. Note that detaching also means the bridge no longer receives signals sent to acpx's process group; both are left to a follow-up that adds a real die-with-parent mechanism.
- **Headless assumption.** `detached` puts the bridge in a new session without a controlling terminal; tools that need `/dev/tty` directly would be affected. acpx bridges are headless ACP adapters, so this is expected.

## Real behavior proof

Deterministic regression (`--exit-on-stdin-end` makes the mock bridge exit on stdin EOF, so only the post-exit sweep can reap the grandchild). Same test, with and without the fix:

```
# WITH the fix
✔ AcpClient close reaps grandchildren when the bridge exits during stdin-close
ℹ pass 1  ℹ fail 0

# WITHOUT the fix (git stash of src/acp/client.ts)
✖ AcpClient close reaps grandchildren when the bridge exits during stdin-close
  AssertionError: grandchild 57370 must be dead after close even when the bridge exited on its own
ℹ pass 0  ℹ fail 1
```

Full suite: `850 tests, 847 pass, 0 fail, 3 skipped`. `typecheck` + `lint` clean.

## Tests

`test/client.test.ts` (+ `test/mock-agent.ts`), real-process, Unix:
- grandchild reaped when the bridge is **stubborn** (`--ignore-sigterm --stay-alive-after-stdin-end`, forcing the group `SIGKILL`);
- grandchild reaped when the bridge **exits cleanly** on stdin-close (`--exit-on-stdin-end`, exercising the post-exit group sweep — the case ClawSweeper flagged).

---

Addresses the ClawSweeper review of the previous revision (the stdin-close descendant leak). Maintainer edits enabled (personal fork).

### DIFF
--- a/src/acp/client.ts
+++ b/src/acp/client.ts
@@ -442,6 +442,7 @@ export class AcpClient {
   private agentStartedAt?: string;
   private lastAgentExit?: AgentExitInfo;
   private lastKnownPid?: number;
+  private agentGroupLeader = false;
   private readonly promptPermissionFailures = new Map<string, PermissionPromptUnavailableError>();
   private readonly pendingConnectionRequests = new Set<PendingConnectionRequest>();
   private readonly modelConfigIds = new Map<string, string>();
@@ -710,11 +711,16 @@ export class AcpClient {
   private async spawnAgentProcess(
     plan: AgentLaunchPlan,
   ): Promise<ChildProcessByStdio<Writable, Readable, Readable>> {
+    const agentGroupLeader = process.platform !== "win32";
     const spawnedChild = spawn(
       plan.spawnCommand,
       plan.args,
-      buildSpawnCommandOptions(plan.spawnCommand, plan.spawnOptions),
+      buildSpawnCommandOptions(plan.spawnCommand, {
+        ...plan.spawnOptions,
+        detached: agentGroupLeader,
+      }),
     ) as ChildProcessByStdio<Writable, Readable, Readable>;
+    this.agentGroupLeader = agentGroupLeader;
     try {
       await waitForSpawn(spawnedChild);
     } catch (error) {
@@ -1369,6 +1375,7 @@ export class AcpClient {
     this.initResult = undefined;
     this.connection = undefined;
     this.agent = undefined;
+    this.agentGroupLeader = false;
   }
 
   private async terminateAgentProcess(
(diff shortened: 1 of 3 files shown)


superWorldSavior:
Updated to address the ClawSweeper review:

- **Fixed the stdin-close descendant leak** — a final best-effort process-group sweep now runs even when the bridge PID exits on its own during the stdin-close grace, so descendants (model process, MCP servers) are still reaped. New deterministic regression (`--exit-on-stdin-end`) is red without the fix / green with it.
- **Error observability** — group/PID signal failures other than `ESRCH` are now traced instead of silently swallowed.
- **Documented the scope explicitly** in the PR body: process *group* (not full tree), POSIX-only, cooperative-close only (abrupt death + re-detached descendants left to a follow-up), headless assumption.
- Real behavior proof (red-without / green-with transcript) added to the PR body; `typecheck` + `lint` clean, full suite green.

@clawsweeper re-review

superWorldSavior:
Superseded by #433, which now carries this commit as its base **and** gates POSIX detachment on a working lifeline helper — so the process-group reaping and the abrupt-death watchdog ship as one coherent, no-regression change. Continuing the review on #433.

PR #433, opened 2026-07-05
feat(acp): pipe lifeline — reap the bridge tree on abrupt owner death

superWorldSavior:
## Problem

`AcpClient.close()` and abrupt queue-owner death both need to clean up the whole ACP bridge process group, not just the bridge PID.

Before this PR, bridge descendants could survive in several cases:

- cooperative close could kill only the bridge process and leave bridge-spawned children alive;
- initialize/startup failure could hit a PID-only `child.kill()` path before the normal close machinery owned the bridge;
- abrupt owner death (`SIGKILL`, OOM kill, crash) runs no JavaScript cleanup at all, so a detached bridge tree can be orphaned under PID 1.

We hit the abrupt-death variant in production with leaked bridge/model/MCP processes accumulating until new sessions hung at `session/new`.

## Fix

This PR makes the bridge lifecycle group-aware on POSIX:

- The bridge is spawned as a detached process-group leader when a POSIX lifeline helper is available.
- Normal `AcpClient.close()` first closes stdin for graceful ACP shutdown, but then checks whether the bridge process group is still alive. If descendants remain, it signals the group with `SIGTERM`, waits, then falls back to `SIGKILL`.
- Initialize/startup failure now also uses group-aware termination instead of PID-only `child.kill()`.
- Abrupt owner death is covered by a tiny native pipe lifeline helper. The helper inherits a pipe from the owner; owner EOF means the owner died, so the helper reaps the bridge process group.

## Lifeline safety model

The lifeline uses pipe identity, not owner PID polling, so owner PID reuse is not part of the death-detection mechanism.

The helper also guards the bridge PGID side:

- strict PGID parsing rejects malformed or unsafe input;
- `kill(-pgid, 0)` probes the process group itself, not just the leader, so TERM-resistant descendants still receive the `SIGKILL` fallback;
- if the watched process group disappears while the owner is still alive, the helper disarms and exits cleanly, avoiding a stale-PGID kill on later owner EOF;
- `POLLERR`/`POLLNVAL` on the lifeline fd degrade without reaping rather than risk killing a live session.

## No binaries shipped

The npm package ships the C source only. Resolution order is:

1. `ACPX_LIFELINE_HELPER` override
2. package-relative prebuilt helper, if present in a source checkout or future platform package
3. lazy one-time compile into `~/.acpx/native/lifeline-<version>-<platform>-<arch>` using `cc`

If no usable helper can be resolved or compiled, acpx degrades to the previous non-detached/no-lifeline behavior instead of adding a hard native dependency.

## Scope and limitations

- POSIX-only. Windows keeps current behavior.
- Process-group cleanup, not a universal process-tree reaper. A descendant that deliberately starts a new session/process group can escape.
- There is still a tiny best-effort window between bridge spawn and watchdog spawn where an immediate owner `SIGKILL` can orphan the bridge.
- If a helper starts and then immediately dies, abrupt-death cleanup is not protected, but normal close still group-kills the bridge descendants.

## Regression coverage

Added real-process tests for:

- initialize failure with a long-lived bridge grandchild;
- normal close with long-lived bridge grandchildren;
- cooperative bridge exit while a TERM-resistant grandchild survives;
- native lifeline `SIGKILL` fallback after the bridge leader exits;
- native lifeline disarm when the watched process group disappears before owner EOF;
- queue-owner `SIGKILL` cleanup with a TERM-resistant descendant.

## Verification

Final HEAD: `6c5a4f0e577ff677786f2b289e7785f59e83846a`

Local validation:

- `pnpm run check` -> pass (`864` runtime tests + `109` flow/runtime coverage tests, 0 failures; coverage gates passed)
- `pnpm run mutate` -> pass (Stryker score `90.55`, threshold `80`, 0 timeouts)
- `.agents/skills/autoreview/scripts/autoreview --mode branch --base upstream/main` -> clean, no accepted/actionable findings
- `pnpm dlx slophammer-ts@latest rules --format text` -> pass
- `pnpm dlx slophammer-ts@latest dry .` -> `DRY candidates: 0; maximum: 0`
- `pnpm dlx slophammer-ts@latest check . --only ts.dependency-boundaries-required` -> pass

GitHub CI on `ubuntu-24.04`:

- Format, Typecheck, Lint, Build, Conformance Smoke, Test, Mutation, Slophammer -> pass
- Docs -> skipped by scope

## Relationship to #432

This PR supersedes #432 and is self-contained: it includes the cooperative close/group cleanup plus the abrupt-owner-death lifeline work.

---

Maintainer edits enabled (personal fork).

### DIFF
--- a/native/lifeline.c
+++ b/native/lifeline.c
@@ -1,0 +1,137 @@
+#include <errno.h>
+#include <limits.h>
+#include <poll.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#define GROUP_POLL_TIMEOUT_MS 100
+
+static int parse_pgid(const char *value, pid_t *pgid) {
+  errno = 0;
+  char *end = NULL;
+  long parsed = strtol(value, &end, 10);
+  if (errno == ERANGE || end == value || *end != '\0' || parsed <= 1) {
+    return -1;
+  }
+
+  pid_t narrowed = (pid_t)parsed;
+  if ((long)narrowed != parsed) {
+    return -1;
+  }
+
+  *pgid = narrowed;
+  return 0;
+}
+
+static int process_group_alive(pid_t pgid) {
+  if (kill(-pgid, 0) == 0) {
+    return 1;
+  }
+  if (errno == EPERM) {
+    return 1;
+  }
+  if (errno == ESRCH) {
+    return 0;
+  }
+  return 1;
+}
+
+static void reap_group(pid_t pgid) {
+  kill(-pgid, SIGTERM);
+  usleep(200000);
+  /*
+   * The process group can outlive its leader. Probe the group itself so
+   * TERM-resistant descendants still get the SIGKILL fallback.
+   */
+  if (kill(-pgid, 0) == 0 || errno == EPERM) {
+    kill(-pgid, SIGKILL);
+  }
+}
+
+static int should_reap_after_owner_eof(pid_t bridge_pgid) {
+  return process_group_alive(bridge_pgid) ? 1 : 0;
+}
+
+static int wait_for_owner_pipe(pid_t bridge_pgid) {
+  struct pollfd pfd = {
+    .fd = STDIN_FILENO,
+    .events = POLLIN,
+  };
+  char buffer[32];
+
+  for (;;) {
+    int ready = poll(&pfd, 1, GROUP_POLL_TIMEOUT_MS);
+    if (ready == -1) {
+      if (errno == EINTR) {
+        continue;
+      }
+      perror("poll");
+      return -1;
+    }
+
+    if (ready == 0) {
+      if (!process_group_alive(bridge_pgid)) {
+        return 0;
+      }
+      continue;
+    }
+
+    if ((pfd.revents & POLLIN) != 0) {
+      ssize_t nread = read(STDIN_FILENO, buffer, sizeof(buffer));
+      if (nread == -1) {
+        if (errno == EINTR) {
+          continue;
+        }
+        perror("read");
+        return -1;
+      }
+      if (nread == 0) {
+        return should_reap_after_owner_eof(bridge_pgid);
+      }
+      if (memchr(buffer, 'R', (size_t)nread) != NULL) {
+        return 0;
+      }
+    }
+
+    if ((pfd.revents & POLLHUP) != 0) {
+      return should_reap_after_owner_eof(bridge_pgid);
+    }
+
+    if ((pfd.revents & (POLLERR | POLLNVAL)) != 0) {
+      /*
+       * The lifeline fd itself is broken: we can no longer observe the owner.
+       * Degrade to pre-lifeline behavior (exit without reaping) rather than
+       * risk killing a live session, and avoid a POLLERR busy-loop.
+       */
+      return -1;
+    }
+
+    if (!process_group_alive(bridge_pgid)) {
+      return 0;
+    }
+  }
+}
+
+int main(int argc, char **argv) {
+  if (argc != 2) {
+    fprintf(stderr, "usage: lifeline <bridgePgid>\n");
+    return 2;
+  }
+
+  pid_t bridge_pgid = 0;
+  if (parse_pgid(argv[1], &bridge_pgid) != 0) {
+    fprintf(stderr, "invalid bridge pgid\n");
+    return 2;
+  }
+
+  int result = wait_for_owner_pipe(bridge_pgid);
+  if (result == 1) {
+    reap_group(bridge_pgid);
+    return 0;
+  }
+  return result == 0 ? 0 : 1;
+}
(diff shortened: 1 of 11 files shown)


superWorldSavior:
Addressed the P1 (detach without a lifeline could regress abrupt group-signal death):

- **POSIX detachment is now gated on a resolved lifeline helper.** `spawnAgentProcess` resolves/compiles the helper *before* the spawn; `agentGroupLeader` (and therefore `detached`, the close-time group reap, and the watchdog) is only enabled when a working helper resolves. If no helper compiles/resolves, the bridge stays **attached** with the pre-lifeline single-PID kill — **zero behavior change**. We only ever detach when a watchdog can reap the group.
- **Regression coverage on both paths**: existing helper-present tests (grandchild reaped on cooperative close + abrupt owner SIGKILL) and a new helper-unavailable test (`ACPX_LIFELINE_HELPER` → invalid path ⇒ bridge stays attached, no watchdog, `close()` still kills the bridge).

Full suite 858/858, `typecheck` + `lint` clean. The native compile/cache surface is unchanged (security was already cleared) — that operational-ownership call remains yours.

@clawsweeper re-review

superWorldSavior:
Reworked the fix after a closer look — the previous `agentGroupLeader = false` on watchdog failure was the wrong lever.

`agentGroupLeader` means **"the bridge was spawned detached"**, not "the watchdog is alive". The bridge is detached the moment a helper resolves and cannot be re-attached post-spawn, so clearing the flag did **not** fall back to an attached bridge — it only disabled the close-time group-kill, leaking the bridge's grandchildren on cooperative close. The new regression test asserts the opposite: a resolved-but-unusable helper **still group-kills the bridge's descendants at close**.

On the "fall back to attached single-PID" suggestion: for the failure this P1 is about — the sole owner dying via a plain SIGKILL / crash / OOM — an attached bridge is orphaned just the same (SIGKILL does not cascade to the group). Attaching would only help a group-wide `kill -9 -<owner-pgid>` or a terminal SIGHUP, which is not the reported case; and the old rollback was not attaching anyway. A real attached fallback would need a probe or re-spawn, which we deliberately skip: the helper is compiled lazily for the host arch and blocks on `poll(stdin)`, so it never fails to run — the failure modes are a misconfigured `ACPX_LIFELINE_HELPER` or a corrupt cache, where degrading to `main`'s no-lifeline behavior is acceptable.

Also hardened:
- Resolution now requires a regular **executable** file (X_OK), not mere existence — a present-but-non-executable helper is unavailable before the detach decision.
- `await waitForSpawn(watchdog)` surfaces spawn errors (ENOENT/EACCES) without an unhandled rejection.

Known best-effort limit (documented in code): a helper that execs then exits immediately (e.g. a corrupt `0755` binary hitting the glibc `execvp -> /bin/sh` fallback) leaves the watchdog dead — close still group-kills, abrupt death is unprotected, same as `main`.

Full suite green (860), typecheck + lint clean.

@clawsweeper re-review

superWorldSavior:
@clawsweeper re-review

Updated the branch after the P1 review. The fix now covers the process-tree lifecycle as three explicit cases:

- initialize/startup failure: `handleInitializeFailure` now uses the same group-aware termination path instead of PID-only `child.kill()` before releasing the lifeline.
- normal cooperative close: `AcpClient.close()` no longer treats bridge exit as sufficient when the detached process group still has descendants; it signals/waits on the group and falls back to SIGKILL.
- abrupt owner death: the native lifeline now kills TERM-resistant descendants, but disarms itself once the watched process group disappears before owner EOF, avoiding stale PGID reuse risk.

Regression coverage added for:

- initialize failure with long-lived bridge grandchild
- cooperative bridge exit with TERM-resistant grandchild
- native lifeline SIGKILL fallback after bridge leader exit
- native lifeline disarm when the process group disappears before owner EOF
- queue-owner SIGKILL lifecycle with TERM-resistant descendant

Verification run on final HEAD `6c5a4f0`:

- `pnpm run check` -> pass (`864` runtime tests + `109` flow/runtime coverage tests, 0 failures; coverage gates passed)
- `pnpm run mutate` -> pass (Stryker score `90.55`, threshold `80`, 0 timeouts)
- `.agents/skills/autoreview/scripts/autoreview --mode branch --base upstream/main` -> clean, no accepted/actionable findings
- `pnpm dlx slophammer-ts@latest rules --format text` -> pass
- `pnpm dlx slophammer-ts@latest dry .` -> `DRY candidates: 0; maximum: 0`
- `pnpm dlx slophammer-ts@latest check . --only ts.dependency-boundaries-required` -> pass

The branch was also squashed down to two commits for a cleaner review history.

superWorldSavior:
@clawsweeper re-review

Final CI state is now fully terminal on HEAD `6c5a4f0`:

- Build, Conformance Smoke, Format, Lint, Slophammer, Test, Typecheck: pass
- Mutation: pass (`6m55s`)
- Docs: skipped by scope

Requesting a final refresh because the previous ClawSweeper review still listed Mutation as pending at review time.

steipete:
Closing: the native-helper cache introduces an untrusted, predictable execution boundary (a foreseeable-path binary that gets executed), and the branch is dirty against main. The pipe-lifeline goal is sound — a variant that keeps the helper inside the package boundary with integrity checks would be worth discussing first in an issue. Thanks @superWorldSavior.

PR #444, opened 2026-07-13
fix(cli): report a cold-start session as "agent starting", not "needs reconnect"

guettlibot:
## What

The prompt banner reports a session with no live queue owner as `agent needs reconnect`. Rename that status to `agent starting`.

```
- [acpx] session calm-forest (abc123) · /home/user/project · agent needs reconnect
+ [acpx] session calm-forest (abc123) · /home/user/project · agent starting
```

## Why

`resolveSessionConnectionStatus` derives the banner status from `probeQueueOwnerHealth`, and renders anything unhealthy as `needs reconnect`:

```ts
type SessionConnectionStatus = "connected" | "needs reconnect";

async function resolveSessionConnectionStatus(record) {
  const health = await probeQueueOwnerHealth(record.acpxRecordId);
  return health.healthy ? "connected" : "needs reconnect";
}
```

But a session without a live queue owner isn't broken — the owner is spawned on demand by the very prompt this banner introduces (`sendSession` → `spawnQueueOwnerProcess`). So **every cold start lands here**: a fresh container, a CI job, a per-task worker Pod. Those runs print `agent needs reconnect` on stderr and then complete perfectly.

The string describes a defect that doesn't exist, at precisely the moment nothing needs reconnecting. And because it's on stderr and reads like an auth/connection failure, it's indistinguishable from a real one to anything reading logs. We run acpx one-shot per task in Kubernetes, so *100%* of our runs emit it; a supervisor of ours pattern-matched the line as a Claude auth failure and parked its whole fleet on healthy credentials. That regex was our bug to own — but the wording is what invited it, and no other acpx output distinguishes "cold start, about to spawn an owner" from "this session is broken".

`starting` states what acpx is about to do. It also matches the tone of the existing sibling string, `session idle; queue owner will start on next prompt`.

## Before / after

Real runs, published `acpx@0.12.0` vs this branch. Both succeed (`stopReason: end_turn`, exit 0):

**Before**
```
$ npx acpx@0.12.0 --approve-all --timeout 90 --format json claude -s demo --file p.md
[acpx] session demo-before (f17003f4-…) · /tmp/acpxdemo · agent needs reconnect
```

**After**
```
$ node dist/cli.js --approve-all --timeout 90 --format json claude -s demo --file p.md
[acpx] session demo-after (da550294-…) · /tmp/acpxdemo · agent starting
```

## Notes

- Behaviour is unchanged; this is the status wording only.
- `SessionConnectionStatus` is internal to `render.ts`. `formatPromptSessionBannerLine` is exported, but its `connectionStatus` parameter is optional and defaults to the cold-start value, so existing callers are unaffected.
- I deliberately left two adjacent things alone to keep this focused, happy to open separate PRs if you'd like them:
  - the banner still goes to **stderr**, and is still printed under `--format json` (it's only suppressed under `--json-strict` / `--format quiet`);
  - `spawnQueueOwnerProcess` uses `stdio: "ignore"`, so when a queue owner *does* fail to start you get `Session queue owner failed to start for session <id>` with no cause. That one cost us a production outage we could not diagnose.

## Testing

- `pnpm run check` passes (format, typecheck, lint, build, viewer, tests + coverage).
- `pnpm run test`: 849/849 pass.
- Updated the two banner assertions in `test/cli.test.ts`, and added one pinning the `connected` wording so the regression can't quietly come back.
- Verified end-to-end against a real `claude` agent (output above).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ Repo: https://github.com/openclaw/acpx
 
 - Runtime/sessions: use collision-resistant temporary paths for concurrent atomic session and index writes. Thanks @henkterharmsel.
 
+- CLI/status: report a normal cold-start session as `agent starting` while preserving `needs reconnect` for an unreachable live owner. Thanks @guettli.
+
 ## 2026.7.23 (v0.12.1)
 
 ### Changes
--- a/src/cli/output/render.ts
+++ b/src/cli/output/render.ts
@@ -16,17 +16,23 @@ function formatRoutedFrom(sessionCwd: string, currentCwd: string): string | unde
   return relative.startsWith(".") ? relative : `.${path.sep}${relative}`;
 }
 
-// A session with no healthy queue owner is not broken: the owner is started on
-// demand by the very prompt this banner introduces. Cold-start every run (a
-// fresh container, a CI job) lands here, so the status has to read as the
-// normal path, not as a defect the operator must act on.
-type SessionConnectionStatus = "connected" | "starting";
+type SessionConnectionStatus = "connected" | "starting" | "needs reconnect";
+
+export function classifySessionConnectionStatus(health: {
+  healthy: boolean;
+  hasLease: boolean;
+}): SessionConnectionStatus {
+  if (health.healthy) {
+    return "connected";
+  }
+  return health.hasLease ? "needs reconnect" : "starting";
+}
 
 async function resolveSessionConnectionStatus(
   record: SessionRecord,
 ): Promise<SessionConnectionStatus> {
   const health = await probeQueueOwnerHealth(record.acpxRecordId);
-  return health.healthy ? "connected" : "starting";
+  return classifySessionConnectionStatus(health);
 }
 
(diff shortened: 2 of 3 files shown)


steipete:
Landed the cold-start banner fix as a three-state rewrite with human contributor credit preserved.

The final behavior distinguishes the actual queue-owner states instead of treating every unhealthy probe alike: healthy owners report `connected`, lease-free cold starts report `starting`, and an unreachable owner that still holds a lease reports `needs reconnect`.

- Exact head: `951c1acd69652b294958a172e2714754db7c0a4e`
- Merge commit: `ac8753bf7f0975d78df349cb0f1417c6631ecbb8`
- Focused proof: banner formatting plus all three health/lease classifications passed
- Attribution: the AI co-author trailer was removed; Thomas Güttler remains the human co-author
- AutoReview: clean at 0.96
- Hosted proof: approved fork run https://github.com/openclaw/acpx/actions/runs/30278623260 passed Test, Mutation, Slophammer, Docs, and the full exact-head matrix

Thanks @guettli and @guettlibot.

PR #447, opened 2026-07-13
fix(session): prevent concurrent save failures

henkterharmsel:
## What Problem This Solves

Fixes an issue where concurrent saves of the same ACPX session could choose the same temporary pathname. One writer could rename the shared temporary file while another writer still expected it to exist, causing an `ENOENT` failure or allowing one save to overwrite another writer's temporary payload.

## Why This Change Was Made

Atomic-write temporary paths now include a per-write UUID in addition to the target path. All session-store callers use the same helper, so concurrent writes retain atomic rename semantics without sharing temporary files.

## User Impact

Concurrent runtime and checkpoint updates can persist the same session reliably without intermittent temporary-file collisions.

## Evidence

- Added a deterministic concurrent-save regression test that holds both writes before rename and verifies distinct temporary paths and a valid final record.
- `pnpm run check` passes on Node 24.18.0 and pnpm 10.33.2.
- 849 regular tests and 110 coverage tests pass; formatting, typecheck, lint, build, viewer checks, and coverage thresholds are green.
- `codex review` and the project autoreview both reported no findings.

## Real behavior proof

**Behavior or issue addressed:** Concurrent saves through the real file session store must not share temporary paths or fail with `ENOENT`.

**Real environment tested:** Linux checkout of this branch using Node 24.18.0 and pnpm 10.33.2. The probe imported the production `createFileSessionStore` implementation directly; it did not use Vitest, mocks, or snapshots.

**Exact steps or command run after this patch:** Created a temporary ACPX state directory, froze `Date.now()` to force every write into the same millisecond, and ran 50 rounds of eight concurrent `store.save(record)` calls. It then loaded the session through the public store, parsed the destination JSON directly, and checked for leftover `.tmp` files.

**Evidence after fix:**

```text
concurrent_saves=400
loaded_session=sid-proof
json_schema=acpx.session.v1
temp_files_remaining=0
result=PASS
```

**Observed result after fix:** All 400 concurrent writes completed without an exception. The final session was readable through the public API, its JSON was valid, and no temporary files remained.

**What was not tested:** I did not interfere with persistence timing on a production gateway, because doing so would risk live session state. Windows and macOS were not tested locally.

## Testing boundary

The race was reproduced deterministically through the real file-session-store implementation with controlled filesystem timing. I did not attempt to force the timing window against a production gateway because that would require interfering with live session persistence.

## AI-assisted disclosure

This PR is AI-assisted. I reviewed the implementation and understand that uniqueness is added only to the temporary pathname; the destination path and atomic rename behavior are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Repo: https://github.com/openclaw/acpx
 
 - Flows: swallow best-effort heartbeat write failures at the timer boundary so storage errors do not become unhandled promise rejections. Thanks @SebTardif.
 
+- Runtime/sessions: use collision-resistant temporary paths for concurrent atomic session and index writes. Thanks @henkterharmsel.
+
 ## 2026.7.23 (v0.12.1)
 
 ### Changes
--- a/src/runtime/public/file-session-store.ts
+++ b/src/runtime/public/file-session-store.ts
@@ -1,6 +1,7 @@
 import fs from "node:fs/promises";
 import path from "node:path";
 import { assertPersistedKeyPolicy } from "../../persisted-key-policy.js";
+import { createAtomicWriteTempPath } from "../../session/persistence/atomic-write.js";
 import { parseSessionRecord } from "../../session/persistence/parse.js";
 import { serializeSessionRecordForDisk } from "../../session/persistence/serialize.js";
 import type { AcpFileSessionStoreOptions, AcpSessionRecord, AcpSessionStore } from "./contract.js";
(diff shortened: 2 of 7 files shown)


henkterharmsel:
@clawsweeper re-review

steipete:
Landed the atomic-write collision fix with contributor credit preserved.

The failure came from using only PID plus millisecond time for temporary filenames: concurrent same-process writes could select the same path, and the first rename removed it from the later writer. The final patch centralizes UUID-suffixed same-directory temp paths across the public runtime store, session repository, and session index.

- Exact head: `2505217507837b3bf1eda6651eb28d56ac13122a`
- Merge commit: `1d17a61f7eb3feda6c66370d1f990c6f4216db51`
- Regression proof: deterministic fixed-ID temp-path test plus concurrent production-store saves
- Focused tests: 12 passed
- Branch AutoReview: clean at 0.99
- Hosted proof: approved fork run https://github.com/openclaw/acpx/actions/runs/30273168889 passed Test, Mutation, Slophammer, Docs, and the full exact-head matrix

Thanks @henkterharmsel.

PR #448, opened 2026-07-14
fix(session): surface queue-owner exit status and stderr on cold start

### LINKED ISSUE #445: Queue owner is spawned with stdio:"ignore", so a fatal startup error is reported as a timeout with no cause
## Summary

`sendSession()` spawns the queue owner detached with `stdio: "ignore"` and never observes the child. If the owner dies during startup, acpx reports the **poll timing out** rather than the actual failure:

```js
function buildQueueOwnerSpawnOptions(payloadFilePath) {
  return { detached: true, stdio: "ignore", env, windowsHide: true };   // owner's stderr is discarded
}
function spawnQueueOwnerProcess(options) {
  spawn(process.execPath, resolveQueueOwnerSpawnArgs(), buildQueueOwnerSpawnOptions(payloadPath)).unref();
}
async function sendSession(options) {
  spawnQueueOwnerProcess(queueOwnerRuntimeOptionsFromSend(options));
  for (let attempt = 0; attempt < QUEUE_OWNER_STARTUP_MAX_ATTEMPTS; attempt += 1) {   // 120 x 50ms
    const queued = await submitToRunningOwner(options, waitForCompletion);
    if (queued) return queued;
    await waitMs(50);
  }
  throw new Error(`Session queue owner failed to start for session ${options.sessionId}`);
}
```

The thrown error carries no exit code, no signal and no output — none of which acpx has, because `.unref()` was called on a child whose stdio went to `/dev/null` and whose `exit`/`error` events are never handled. The user is told the owner "failed to start", which is indistinguishable from a slow start, a crash, a config error, or a permissions problem.

## How we hit it

We run acpx in Kubernetes with `$HOME` on a shared volume, and mounted an emptyDir over `$HOME/.acpx/queues` to keep queue-owner locks Pod-local. Kubernetes creates an emptyDir **owned by root**; our worker runs as uid 1000. The directory is world-writable, so acpx can create and delete lock files in it — but `ensureQueueDir()` does:

```js
await fs.mkdir(baseDir, {recursive: true, mode: 0o700});
await fs.chmod(baseDir, 0o700);     // EPERM: chmod is the owner's privilege alone
```

The owner therefore died on its first syscall, every time, and every run failed with `Session queue owner failed to start`. Because the child was mute, diagnosing it took a full day of process archaeology across ~10 hours of downtime — the kernel had said `EPERM` plainly, to a process nobody was listening to.

(The mount was our doing and we have fixed it on our side. This issue is only about the diagnosability, which would have turned that day into a minute.)

## Suggested fix

Any of these would be sufficient; the first two are the smallest:

1. **Attach the child's failure to the thrown error.** Keep a reference to the spawned process and register `error` / `exit` handlers before polling. If the owner exits before the socket appears, throw with its exit code and signal:
   `Session queue owner exited with code 1 before binding its socket`.
2. **Give the owner somewhere to speak.** Instead of `stdio: "ignore"`, direct stderr to a file next to the lock (e.g. `$HOME/.acpx/queues/<key>.owner.log`, or `stdio: ["ignore", "ignore", fd]`), and quote its tail in the error. Keeps the child detached, costs one fd.
3. **Fail loudly on `ensureQueueDir()`.** A queue directory that cannot be prepared is a hard, actionable configuration error; it deserves its own message naming the path, the failing operation and the errno, rather than surfacing later as a startup timeout.

Happy to send a PR for (1) + (2) if you'd like the shape confirmed first.

## Versions

acpx 0.10.0 and 0.12.0 (identical behaviour on both), Node 22, Linux/amd64.

SebTardif:
Closes #445

## What Problem This Solves

Fixes an issue where operators running acpx would see a generic
`Session queue owner failed to start` timeout when the detached queue
owner died during cold start (for example `chmod EPERM` on a root-owned
emptyDir-mounted queue directory). The real exit code, signal, and
stderr were discarded because the owner was spawned with `stdio: "ignore"`
and no `exit`/`error` listeners, so diagnosing the failure required
process archaeology instead of a clear error message.

## Why This Change Was Made

Keep the owner detached. During cold start only, pipe stderr into a
byte-capped in-memory buffer, observe exit/spawn errors while polling
for the socket, and include that evidence in the startup failure. After
the owner binds (or after failure), drop the pipe so a long-lived owner
(including `--ttl 0`) cannot retain unbounded diagnostics. Also make
`ensureQueueDir` report the path and errno when mkdir/chmod fails.

## User Impact

When a queue owner dies before binding, the error names the session,
exit code (or spawn error), and a bounded stderr tail, instead of looking
like a slow start. Successful starts do not keep capturing owner stderr
for the process lifetime.

## Evidence

Early-exit diagnostics (after bound-capture fix):

```console
$ node --input-type=module <<'JS'
// force owner to write stderr and exit 7
process.env.ACPX_QUEUE_OWNER_ARGS = JSON.stringify([
  "-e", "console.error('boom EPERM chmod'); process.exit(7)",
]);
// spawnQueueOwnerProcess + formatQueueOwnerStartupFailure ...
Session queue owner failed to start for session proof-sess: exited with code 7 before binding its socket: stderr:
boom EPERM chmod
exit {"exited":true,"code":7,"signal":null}
default stdio: ignore
capture stdio: [ 'ignore', 'ignore', 'pipe' ]
```

Lifecycle tests:
- `captures early stderr then stops after stopStartupCapture` (post-bind noise not retained)
- `bounds captured stderr to QUEUE_OWNER_STARTUP_STDERR_MAX_BYTES` (20k write capped)

`pnpm test` → 854 pass; lint/typecheck clean.

## Real behavior proof

- **Behavior or issue addressed:** Cold-start queue-owner death used to surface only as a poll timeout with no exit code or stderr; unbounded file capture was unsafe for long-lived owners.
- **Real environment tested:** macOS darwin arm64, Node v26.5.0, branch `fix/queue-owner-startup-diagnostics`.
- **Exact steps or command run after this patch:** Forced `ACPX_QUEUE_OWNER_ARGS` early-exit process; unit tests for stop-capture and byte bound.
- **Evidence after fix:** Message includes exit code 7 + stderr; capture freezes after `stopStartupCapture`; buffer capped at 4KB.
- **Observed result after fix:** Failures are attributable; successful bind drops the stderr pipe.
- **What was not tested:** Full multi-agent production send against a live ACP backend in Kubernetes; Windows-specific pipe destroy edge cases in production.

## Summary

- Bounded in-memory stderr capture during cold start only
- `stopStartupCapture()` after successful submit or terminal failure
- Clearer ensureQueueDir mkdir/chmod errors
- Tests for formatter, pipe options, capture stop, and byte bound

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Repo: https://github.com/openclaw/acpx
 
 ### Fixes
 
+- Session queue owner: capture owner stderr and exit status during cold start so a dead owner reports the real failure instead of a silent timeout. Thanks @SebTardif.
+
 ## 2026.7.4 (v0.12.0)
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Repo: https://github.com/openclaw/acpx
 
 ### Fixes
 
-- Session queue owner: capture a bounded owner stderr tail and exit status during cold start only (pipe dropped after bind) so a dead owner reports the real failure instead of a silent timeout. Thanks @SebTardif.
+- Session queue owner: capture a bounded owner stderr tail and exit status during cold start only (stop retaining at first IPC accept; keep draining the pipe so long-lived owners are not killed by EPIPE) so a dead owner reports the real failure instead of a silent timeout. Thanks @SebTardif.
 
 ## 2026.7.4 (v0.12.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Repo: https://github.com/openclaw/acpx
 
 ### Fixes
 
-- Session queue owner: capture owner stderr and exit status during cold start so a dead owner reports the real failure instead of a silent timeout. Thanks @SebTardif.
+- Session queue owner: capture a bounded owner stderr tail and exit status during cold start only (pipe dropped after bind) so a dead owner reports the real failure instead of a silent timeout. Thanks @SebTardif.
 
 ## 2026.7.4 (v0.12.0)
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -3,20 +3,11 @@
 import { realpathSync } from "node:fs";
 import { fileURLToPath, pathToFileURL } from "node:url";
 import { main } from "./cli-core.js";
+import { installBrokenPipeHandler } from "./cli/broken-pipe.js";
 import { buildQueueOwnerArgOverride } from "./cli/session/queue-owner-process.js";
 
 export { formatPromptSessionBannerLine } from "./cli-core.js";
 export { parseAllowedTools, parseMaxTurns, parseTtlSeconds } from "./cli/flags.js";
-
-function installBrokenPipeHandler(stream: NodeJS.WritableStream): void {
-  stream.on("error", (error: NodeJS.ErrnoException) => {
-    if (error.code === "EPIPE") {
-      process.exit(0);
-    }
-
-    throw error;
-  });
-}
 
 function isCliEntrypoint(argv: string[]): boolean {
   const entry = argv[1];
@@ -35,8 +26,11 @@ function isCliEntrypoint(argv: string[]): boolean {
 }
 
 if (isCliEntrypoint(process.argv)) {
-  installBrokenPipeHandler(process.stdout);
-  installBrokenPipeHandler(process.stderr);
+  const isQueueOwner = process.argv[2] === "__queue-owner";
+  installBrokenPipeHandler(process.stdout, "exit");
+  // After the submitting CLI exits, a detached owner loses its stderr reader.
+  // Ignore that expected EPIPE so later diagnostics cannot kill the owner.
+  installBrokenPipeHandler(process.stderr, isQueueOwner ? "ignore" : "exit");
 
   const queueOwnerArgOverride = buildQueueOwnerArgOverride(fileURLToPath(import.meta.url));
   if (queueOwnerArgOverride) {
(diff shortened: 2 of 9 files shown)


SebTardif:
@clawsweeper re-review

Addressed the P1: stderr is now a startup-only piped buffer (4KB max), destroyed via stopStartupCapture after bind or failure. No full-lifetime file retention. Added lifecycle and bound tests.

SebTardif:
@clawsweeper re-review

Addressed both P1s on top of the bounded-capture work:

1. **IPC accept boundary:** `onQueueAccepted` stops retained capture when the owner first acknowledges the submit request (not after full turn completion when `waitForCompletion` is true).
2. **Drain not destroy:** `stopStartupCapture` keeps the stderr pipe open and discards further data so post-bind writes cannot EPIPE-kill a long-lived owner.

Tests cover stop-retain + clean exit after post-bind stderr writes, and the 4KB bound.

steipete:
Merged after maintainer repair on exact head `57caa004b8838c853393777ac0ed749f10934ff1`.

Proof:

- `pnpm run check`: green locally, including 860 tests and the separate 109-test coverage run (94.14% statements, 87.36% branches).
- AutoReview: clean on both the local repair and the full `origin/main...HEAD` branch candidate.
- GitHub Actions run 29688387168: scope, docs, lint, format, typecheck, conformance smoke, Slophammer, test, build, and mutation jobs all green.
- Regression coverage includes early-exit status/stderr, bounded diagnostic tail retention, clean lease-race exits, detached parent liveness, and queue-owner stderr EPIPE handling.
- Model-identifier audit: PASS; the candidate introduces no provider model identifiers.

The original contributor commits and changelog credit for @SebTardif remain in the PR history. Squash merge: `24ddd69801133691329e96b827beeec0b9015802`.

PR #459, opened 2026-07-18
fix(cli): preserve positive timer durations

realmehmetali:
## Summary

- Preserve positive sub-millisecond `--timeout` and `--ttl` inputs as a 1 ms timer instead of rounding them to the disabled/keep-alive-forever sentinel.
- Reject durations above Node's 2,147,483,647 ms timer ceiling instead of allowing Node to clamp them to an unintended 1 ms timer.
- Add parser regression coverage and an Unreleased changelog entry.

## Root cause

The CLI validated the duration in seconds, then rounded to milliseconds without validating the converted value. A small positive value such as `0.0001` became `0`; downstream, `timeoutMs <= 0` disables the timeout and `ttlMs === 0` means keep alive forever. Oversized finite values also exceeded Node's supported timer range and could be clamped by the runtime.

## Why this matters for frontier AI agents

Frontier AI agent systems depend on precise cancellation and lifecycle semantics. A timeout that silently turns itself off, or an idle TTL that unexpectedly becomes infinite, can strand orchestration capacity and make long-running autonomous workflows difficult to reason about. Keeping these boundary values explicit makes agent infrastructure safer and more predictable.

## Validation

- Required runtime: Node 22.22.0 with pnpm 10.33.2.
- Formatting, typecheck, lint, build, and viewer checks passed.
- `node --test --test-concurrency=4 dist-test/test/*.test.js` — 860 passed, 0 failed.
- Coverage gate — 109 passed; 94.14% statements, 87.36% branches, 96.07% functions, and 94.14% lines.
- Focused parser, lifecycle, compare-command, and built-CLI checks passed.

## Runtime behavior proof

The branch is rebased onto current `origin/main` (`24ddd69`).

- A built CLI run with `--timeout 0.0001` against a mock agent sleeping for 50 ms now exits 3 with `{"jsonrpc":"2.0","id":null,"error":{"code":-32070,"message":"Timed out after 1ms","data":{"acpxCode":"TIMEOUT","origin":"cli","sessionId":"unknown"}}}`.
- The integration lifecycle test now uses `--ttl 0.0001`; the queue owner exits and the session remains resumable. Before this fix, rounding to the zero sentinel would keep the owner alive forever.
- `--timeout 2147483.648` and `--ttl 2147483.648` are explicitly rejected with exit 2 instead of reaching the Node timer clamp.
- Compatibility note for maintainer confirmation: the oversized-value behavior deliberately changes from an unintended near-immediate Node clamp to a clear CLI validation error.

## AI and provenance disclosure

This contribution was prepared with significant AI assistance from OpenAI Codex and was manually scoped, tested, and reviewed. It is based entirely on public repository sources and synthetic test inputs. It contains no private or proprietary code, data, plans, or product-specific examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ Repo: https://github.com/openclaw/acpx
 
 - Session queue owner: capture a bounded owner stderr tail and exit status during cold start only (stop retaining at first IPC accept; keep draining the pipe so long-lived owners are not killed by EPIPE) so a dead owner reports the real failure instead of a silent timeout. Thanks @SebTardif.
 
+- CLI/timers: preserve tiny positive `--timeout` and `--ttl` values as 1 ms instead of disabling timers, and reject delays beyond Node's supported timer range. Thanks @realmehmetali.
+
 ## 2026.7.4 (v0.12.0)
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Repo: https://github.com/openclaw/acpx
 
 - CLI/ACP: add `--no-fs` to disable advertised ACP file read/write capabilities so compatible agents can use their native filesystem implementation. Thanks @zgxkbtl.
 
+- CLI/timers: preserve tiny positive timeout and TTL values from flags or config as 1 ms instead of disabling timers, and reject delays beyond Node's supported timer range. Thanks @realmehmetali.
+
 ### Breaking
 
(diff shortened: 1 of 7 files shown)


steipete:
Landed the timer-boundary fix with contributor credit preserved.

The original flag fix was correct but incomplete because JSON config had a separate seconds-to-milliseconds conversion. The final patch centralizes that ownership boundary for flags and config, preserves only the explicit zero-TTL sentinel, floors positive sub-millisecond values to 1 ms, rejects values beyond Node's timer range, and keeps `config show` round-trippable.

- Exact head: `e2a4ca0e033a043c8a049904659ac2fb84cd60b5`
- Merge commit: `69c20c2bb179cf5ced4376eac1c18bdefd580f91`
- Focused proof: 43 flag/config tests plus the real queue-owner TTL integration passed
- AutoReview: found and drove the fractional TTL display fix; final branch review clean at 0.97
- Hosted proof: approved fork run https://github.com/openclaw/acpx/actions/runs/30276739380 passed Test, Mutation, Slophammer, Docs, and the full exact-head matrix

Thanks @realmehmetali.

PR #469, opened 2026-07-22
feat: add filesystem capability opt-out

zgxkbtl:
## What Problem This Solves

When an ACP client advertises `clientCapabilities.fs`, Qwen delegates file reads and writes to the client. acpx intentionally confines those proxied operations to its configured cwd, so Qwen cannot use delegated file tools for its own runtime temporary directories outside that subtree.

The existing `--no-terminal` flag, introduced by [PR #155](https://github.com/openclaw/acpx/pull/155), provides an opt-out for terminal delegation but there was no equivalent filesystem opt-out.

## Why This Change Was Made

- Add one negatable global flag, `--no-fs`, matching the ACP capability name and the existing `--no-terminal` convention.
- Advertise both `clientCapabilities.fs.readTextFile` and `writeTextFile` as `false` during initialization. Explicit `false` is required for compatible agents such as Qwen to select their native filesystem implementation.
- Preserve the current default (`true`) and keep filesystem and terminal delegation independently configurable.
- Propagate the option through direct prompt/exec, persistent sessions, queue owners, compare runs, and ACP flow nodes.
- Document the generic flag in the landing/CLI/skill docs and the Qwen-specific fallback behavior in `agents/Qwen.md`.

## User Impact

Users can now run:

```bash
acpx --no-fs qwen exec 'inspect the runtime artifact and summarize it'
```

Qwen will use its own file tools instead of issuing `fs/read_text_file` and `fs/write_text_file` requests to acpx. Existing commands retain the current proxied filesystem behavior unless the flag is supplied. `--no-fs` only affects new ACP client initialization and can be combined with `--no-terminal`.

## Evidence

- Reproduced with Qwen Code 0.20.0 from a temporary cwd before the change: the handshake advertised filesystem read/write as `true`; Qwen sent three `fs/write_text_file` requests; acpx rejected the resolved runtime path with `-32603 Path is outside allowed cwd subtree`; no file was written.
- Repeated the same live Qwen scenario with `--no-fs --no-terminal`: the handshake advertised both filesystem capabilities as `false`; Qwen emitted builtin `WriteFile` and `ReadFile` tool updates; the ACP filesystem request count was zero; and the native round trip returned the exact expected file content.
- Added integration coverage for direct `exec --no-fs` and `flow run --no-fs`, plus CLI parsing/routing and queue-owner propagation tests.
- `pnpm run check` passed with Node test concurrency set to 1 locally; coverage remained above policy gates (94.14% statements/lines, 87.36% branches, 96.07% functions).
- `pnpm run check:docs` passed.
- `pnpm run mutate` passed at 90.37% mutation score.
- `pnpm dlx slophammer-ts@latest rules --format text`, `dry .`, and the direct dependency-boundary check passed.
- Repo-local autoreview reported no accepted/actionable findings.


steipete:
Landed the filesystem-capability opt-out with contributor credit preserved.

`--no-fs` now consistently propagates through direct exec, persistent sessions, queue-owner startup, compare, flows, runtime reconnects, and ACP initialize. The default remains unchanged; compatible agents can use their native filesystem path when both advertised client filesystem capabilities are false.

- Exact head: `74688ecd97f341e5916531342ebc26a3b06350d6`
- Merge commit: `2f9fd039463ab255e349ee3b718c1186bfbaa0c5`
- Repair: added compare documentation and moved the attributed changelog entry into current Unreleased
- Focused proof: both direct exec and flow `--no-fs` integration paths passed
- Static/docs gates and AutoReview: clean; AutoReview 0.96
- Hosted proof: approved fork run https://github.com/openclaw/acpx/actions/runs/30274848543 passed Test, Mutation, Slophammer, Docs, and the full exact-head matrix

Thanks @zgxkbtl.

PR #470, opened 2026-07-24
Add Junie as a built-in agent (`junie --acp true`)

mashan555:
feat: add JetBrains Junie as a built-in agent

What Problem This Solves
Users who work with JetBrains Junie could not target it directly through acpx — there was no junie built-in agent, so acpx junie ... did not resolve to a known harness. This change adds Junie to the built-in agent registry so it is available as a first-class native agent alongside the other supported harnesses in the acpx CLI.

Why This Change Was Made
Junie already ships a compliant ACP stdio mode junie --acp true and its ACP adapter correctly advertises persistent-session capabilities loadSession + sessionCapabilities.resume, so it works over the standard native ACP path with no adapter package and no agent-specific glue in src/acp/client.ts. The change is intentionally minimal: register the launch command and update the documentation surfaces required by repo policy. Non-goal: no changes to docs/CLI.md main-landing docs stay impartial and only name pi, openclaw, codex, claude, and no adapter version specifiers in docs.

User Impact
Users can now run Junie directly through acpx, for example:

acpx junie exec 'echo hello'
acpx junie sessions new --name docs
acpx junie -s docs 'rewrite CLI docs'

Named sessions reopen across processes persistent mode because acpx detects Junie's capabilities automatically. Junie must be installed and authenticated locally — acpx never installs it. The one intentional limitation is that exec/prompt without a named session is one-shot by design.

Evidence
• test/agent-registry.test.ts: resolution junie, JUNIE, whitespace, "native, not via adapter", and presence in listBuiltInAgents.
• test/integration.test.ts: writeFakeJunieAgent helper + e2e test that acpx junie exec 'echo hello' resolves to junie --acp true and returns exit code 0 with hello output.
• pnpm run check format, typecheck, lint, build, viewer build, coverage tests passes; pnpm run check:docs passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Repo: https://github.com/openclaw/acpx
 ### Changes
 
 - Agents/built-ins: refresh the default Pi, Codex, Claude, and Mux adapter ranges. Thanks @kelvinschen and @TheAngryPit.
+- Agents/built-ins: add JetBrains Junie as a built-in agent via `junie --acp true`.
 
 ### Breaking
 
--- a/README.md
+++ b/README.md
@@ -359,6 +359,7 @@ Built-ins:
 | `fast-agent` | `uvx fast-agent-mcp acp`                                                    | [fast-agent](https://fast-agent.ai)                                                                             |
 | `grok-build` | native (`grok agent stdio`)                                                 | [Grok Build](https://docs.x.ai/build/overview)                                                                  |
 | `iflow`      | native (`iflow --experimental-acp`)                                         | [iFlow CLI](https://github.com/iflow-ai/iflow-cli)                                                              |
+| `junie`      | native (`junie --acp true`)                                                 | [JetBrains Junie CLI](https://www.jetbrains.com/junie/)                                                         |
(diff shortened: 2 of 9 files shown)


steipete:
Closing this implementation after a full current-main review.

The Junie ACP command itself is credible: JetBrains documents `junie --acp true` over stdio. The submitted branch does not prove the contract it asks acpx to own, though:

- the fake Junie executable accepts missing or malformed ACP arguments, so the integration still passes if acpx launches the wrong command;
- there is no authenticated Junie run or cross-process named-session resume proof;
- the stronger session/load and resume claims are not established by the linked vendor material;
- the head has no project CI; and
- reserving `junie` changes the existing raw-command fallback without an approved compatibility decision.

A design-first follow-up is welcome. It should first define the permanent keyword/override contract, then bring strict argv assertions, real one-shot plus cross-process resume evidence, the official Junie CLI/ACP links, synchronized harness docs, conventional title/changelog attribution, and exact-head CI. Please preserve @mashan555's contribution context if that work proceeds.

Reviewed head: facd630161e6b02994adf3d49b89dbf587c1716b.

PR #478, opened 2026-07-28
fix: harden adapter and queue-owner lifecycles

trumpyla:
## Summary

This fixes two related lifecycle/concurrency failure classes:

- package-exec wrappers could exit while adapter descendants remained alive
- concurrent cold prompts could observe a queue-owner lease before its socket was bound, or mutate the same lease concurrently

The implementation now shares one process-tree supervisor between ACP adapters and terminal commands, publishes and mutates queue-owner leases atomically, serializes lease mutations with crash-recoverable claims, and keeps the optimized startup probe aligned with the full startup grace period.

### Adapter descendants survived wrapper exit

The old client cleanup primarily tracked the direct adapter child. A launch chain such as `npx -> adapter -> resistant grandchild` can outlive that wrapper, change process group, or become reparented before cleanup discovers it.

The shared process-tree implementation now:

- creates an owned process group for adapter launches on POSIX
- captures descendants before signaling and tracks multi-stage wrapper launches for a bounded 10-second window
- uses Linux `/proc` start times as process identities and a fixed-locale POSIX fallback
- validates remembered identities before signaling after root exit
- rescans the owned group and identity-validated escaped descendants during shutdown
- never signals a potentially recycled POSIX process-group ID after its root exits
- rejects a recycled live group leader when its identity differs from the captured root
- applies the final signal to already-tracked live descendants as well as newly discovered descendants
- uses creation-time-validated Windows parent edges, a Windows-specific five-second discovery bound, and `taskkill /t`

The large deletion in `terminal-manager.ts` is deliberate deduplication: its private process-tree implementation moved to `src/acp/process-tree.ts`, which both the ACP client and terminal manager now use.

### Queue-owner lease races

The old queue-owner record and cleanup paths allowed several time-of-check/time-of-use races:

- readers could observe a partially written lease
- refresh, release, stale cleanup, and successor publication could overlap
- stale cleanup could kill an owner that refreshed before cleanup acquired its claim
- restoring a displaced stale claim could overwrite a concurrently acquired claim
- a fresh lease-before-bind owner could be reported as not accepting requests

The lease store now:

- publishes records with same-directory atomic create/replace operations
- assigns collision-resistant generations
- serializes refresh, release, and cleanup with inode-qualified claims
- recovers claims only after PID/process-identity validation
- revalidates owner generation, liveness, and heartbeat while holding the cleanup claim
- restores displaced claims with no-overwrite hard links
- lets a released claim remove its temporarily quarantined record instead of restoring an orphan
- retries transient same-generation release contention and resets failed release state for an in-process retry
- lets a shutting-down owner defer removal to a live external cleanup claimant without a false non-zero exit
- makes unresolved explicit cleanup fail visibly
- treats an already-removed contended lease as successful cleanup
- reacquires stale collisions in the same attempt with fresh timestamps

## Performance

The startup-probe optimization removes nested retry work from every outer startup-loop iteration.

| Unavailable-socket startup probe | Before | This PR | Change |
| --- | ---: | ---: | ---: |
| Socket attempts | 42 (40 submit + 2 health) | 1 | 97.6% fewer |
| Scheduled inner retry delay | 2,100 ms | 0 ms | 2,100 ms removed |
| Redundant health reconnect | yes | no | removed |

The outer loop now derives its budget from the declared startup grace:

- startup grace: **10,000 ms**
- retry interval: **50 ms**
- attempts: **201**
- covered retry window: **10,000 ms**

This preserves the full lease-before-bind window while keeping each fresh-owner probe single-attempt. Established owners retain the prior retry budget so one transient connection refusal does not become a user-visible failure.

### Before/after microbenchmark

Environment: Darwin arm64, Node v26.5.0, five runs against a nonexistent Unix socket.

| Revision | Samples (ms, sorted) | Median |
| --- | --- | ---: |
| `upstream/main` (`504040f`) | 2146.019, 2146.883, 2147.834, 2149.344, 2155.274 | **2147.834 ms** |
| optimized path (`f68abe0`, unchanged in current head) | 0.087, 0.091, 0.094, 0.129, 1.725 | **0.094 ms** |

That is approximately **22,849x lower probe latency (99.996%)**. This isolates the lease-before-bind probe; it is not a general request-throughput benchmark. A regression also asserts that the `queue.connect` metric records exactly one failed connection.

## Regression coverage

The added tests cover:

- normal and failed adapter startup through real `npx` wrapper/grandchild chains
- SIGTERM-resistant adapter descendants
- descendants that detach into another POSIX process group
- final signaling of already-tracked and newly discovered descendants
- POSIX PID/PGID reuse and process-identity changes
- bounded Windows multi-stage descendant tracking
- Windows process-list startup exceeding one second
- bounded process-list command failures
- Linux procfs identity discovery
- atomic lease publication under concurrent readers
- stale malformed and dangling-symlink lock handling
- refresh/release/cleanup claim serialization across processes
- claimant crash recovery
- release while a fresh claim is temporarily quarantined
- same-object release retry after transient claim contention
- concurrent lease removal during explicit cleanup
- cleanup revalidation after a concurrent heartbeat refresh
- no-overwrite restoration when a concurrent claim wins
- one-attempt startup probing and full-grace outer retry coverage
- synchronous detached-test-owner PID publication

## Validation

| Check | Result |
| --- | --- |
| Format, typecheck, type-aware lint, distributable build | passed on `5d3063b` |
| Earlier clean full main suite | 927 passed, 1 platform skip |
| Final full main-suite retry | runner stalled under system load; affected lease test passed immediately in isolation |
| Timed-out npx grandchild test, isolated | passed in 5.06 s |
| Timed-out flow-disconnect test, isolated | passed in 565 ms |
| Coverage-target suite | 111/111 passed |
| Coverage | 94.16% lines/statements, 87.36% branches, 96.07% functions |
| Final affected process-tree/queue suites | 69 passed, 1 platform skip |
| `pnpm run check:docs` | passed |
| `git diff --check` | passed |
| GitHub CI on `5d3063b` | 11 passed, 0 failed, including Test and Mutation |

The prior full run completed cleanly. The final retry later stalled in the unrelated compare suite under unusually high local load and was interrupted; the only implicated lease test passed immediately in isolation. GitHub CI remains the independent clean-run gate for the pushed head.

## Adversarial review

The repository autoreview loop found and drove fixes for:

- stale cleanup without claim-time revalidation
- POSIX descendants missed before unexpected root exit
- final cleanup skipping already-tracked live descendants
- a startup retry window shorter than its declared grace
- claim restoration overwriting a concurrent winner
- release during temporary claim quarantine restoring an orphan
- recycled POSIX groups being adopted after root identity mismatch
- established owners losing transient-connect retries
- Windows process discovery using a one-second PowerShell budget
- release and explicit cleanup abandoning transient claim contention
- failed release promises preventing same-object retry
- claim-driven owner shutdown reporting a false busy failure

The final stable-snapshot Claude `claude-fable-5` autoreview ran on `5d3063b` with tools disabled and reported:

> no accepted/actionable findings reported
>
> overall: patch is correct

No stale pre-fix vendor verdict is claimed for the current head.

## Size and rationale

- **13 commits**
- **17 files**
- **3,138 insertions / 440 deletions**
- production `src/**`: **1,507 insertions / 431 deletions**

Most added lines are the shared cross-platform process-tree implementation, atomic lease/claim handling, and adversarial regression coverage. The terminal-specific deletion removes duplicate lifecycle logic rather than removing behavior.

SBOM, provenance, Homebrew, and SEA packaging remain intentionally separate from this runtime correctness/performance PR.

## Commits

- `e7efc14` — terminate ACP adapter process groups
- `93b6be1` — retry queue-owner lease-before-bind races
- `f68abe0` — harden lifecycle race handling
- `f158b26` — wait for terminal output drain
- `fdf12ea` — publish queue owner leases atomically
- `006e333` — close lifecycle review races
- `2a4e074` — validate remembered process identities
- `2537ffb` — rescan process groups during shutdown
- `79457ae` — close remaining process tree races
- `a85da81` — serialize queue lease mutations
- `ec048e5` — close adapter lifecycle races
- `475bdb8` — preserve concurrent queue lease claims
- `5d3063b` — close adversarial lifecycle races

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,8 @@ Repo: https://github.com/openclaw/acpx
 
 - CLI/status: report a normal cold-start session as `agent starting` while preserving `needs reconnect` for an unreachable live owner. Thanks @guettli.
 
+- Runtime/agents: terminate the owned adapter process group/tree during normal and failed startup cleanup so package-exec wrappers cannot leave descendants running after ACPX exits.
+
 ## 2026.7.23 (v0.12.1)
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,9 @@ Repo: https://github.com/openclaw/acpx
   best-effort tracked process trees on Windows during normal and failed startup
   cleanup so package-exec wrappers do not leave captured descendants running
   after ACPX exits; exit-triggered snapshots are observed before cleanup
-  completes, owned POSIX group members created during shutdown are discovered,
-  external process-list discovery is bounded, and remembered descendants are
+  completes, owned POSIX group members created during shutdown are discovered
+  and re-signaled during forced cleanup, external process-list discovery is
+  bounded, and Windows parent edges plus remembered descendants are
   identity-checked before later signaling.
 
 - Runtime/queue: publish queue-owner leases atomically, preserve fresh malformed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,9 @@ Repo: https://github.com/openclaw/acpx
   best-effort tracked process trees on Windows during normal and failed startup
   cleanup so package-exec wrappers do not leave captured descendants running
   after ACPX exits; exit-triggered snapshots are observed before cleanup
-  completes, external process-list discovery is bounded, and remembered
-  descendants are identity-checked before later signaling.
+  completes, owned POSIX group members created during shutdown are discovered,
+  external process-list discovery is bounded, and remembered descendants are
+  identity-checked before later signaling.
 
 - Runtime/queue: publish queue-owner leases atomically, preserve fresh malformed
   locks during collisions, serialize generation-checked refresh and release,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,17 +16,25 @@ Repo: https://github.com/openclaw/acpx
   best-effort tracked process trees on Windows during normal and failed startup
   cleanup so package-exec wrappers do not leave captured descendants running
   after ACPX exits; exit-triggered snapshots are observed before cleanup
-  completes, owned POSIX group members created during shutdown are discovered
-  and re-signaled during forced cleanup, external process-list discovery is
-  bounded, and Windows parent edges plus remembered descendants are
-  identity-checked before later signaling.
+  completes, POSIX descendants are captured before signaling so later session
+  or group changes cannot escape forced cleanup, owned POSIX group members
+  created during shutdown are discovered and re-signaled, external process-list
+  discovery is bounded, Linux process identities come from procfs, other POSIX
+  snapshots use a fixed locale, Windows wrappers are sampled with bounded
+  exponential backoff through multi-stage launches, transient discovery
+  failures preserve and still signal captured targets, and Windows parent edges
+  plus remembered descendants are identity-checked before later signaling.
 
 - Runtime/queue: publish queue-owner leases atomically, preserve fresh malformed
-  locks during collisions, serialize generation-checked refresh and release,
-  and treat only a fresh lease-before-bind owner as a fast startup miss,
-  preventing premature `QUEUE_NOT_ACCEPTING_REQUESTS` errors without masking
-  older unreachable owners, letting released owners overwrite successors, or
-  leaving stale dangling-symlink locks unrecoverable.
+  locks during collisions, serialize cross-process refresh, release, and stale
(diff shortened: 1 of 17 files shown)


steipete:
Closing: a 3,138-line lifecycle rewrite is too large to land safely, and two P1 defects remain (the Windows fast-wrapper path and recycled-PID handling). If you want to pursue this, splitting it into reviewable slices — starting with the PID-identity model — would give it a real path in. Thanks @trumpyla.

PR #485, opened 2026-08-01
feat: configurable model per agent and per flow node

skywills:
## What

Lets a model be configured per agent and pinned per flow step, instead of only via the global `--model` flag.

- `agents.<name>.model` in config — the default model for sessions started with that agent.
- `model` on flow `acp` nodes — a per-step pin.

Precedence differs by surface, deliberately:

- **CLI** — an explicit `--model` beats `agents.<name>.model`. The flag is a per-invocation instruction the user just typed, so config must not silently take it back.
- **Flows** — node pin → the resolved agent's configured model → the run-wide `--model`. Inside a flow, `--model` is a default for the whole run, so a node or agent pin is the more specific signal.

The configured model seeds a session at **creation** only. It is never reapplied to a live session, so a model chosen with `set model` survives later prompts.

## Why

Running several agents with different models means repeating `--model` on every invocation, and a flow has no way to run a cheap classification step and an expensive implementation step on different models — every ACP step in a run shares one model.

## Changes

| Commit | |
|---|---|
| `feat(config)` | parse and validate `agents.<name>.model` |
| `feat(cli)` | surface it from `resolveAgentInvocation` |
| `feat(flows)` | accept `model` on acp nodes |
| `feat(flows)` | apply node → agent → global precedence via `mergeSessionModel` |
| `feat(cli)` | use the configured model when `--model` is absent |
| `docs` | `docs/config.md`, `docs/flows.md`, `docs/CLI.md`, `skills/acpx/SKILL.md`, CHANGELOG |
| `fix(flows)` | reject a whitespace-only node model |
| `fix(flows)` | record a node model in the flow definition snapshot |
| `fix(cli)` | apply a configured model only at session creation |
| `fix(session)` | keep a configured model off resumed sessions |
| `fix(flows)` | reject conflicting model pins on a shared session |
| `fix(flows)` | keep a configured agent model creation-only in flows |
| `fix(acp)` | send a configured agent model as session-creation metadata |
| `fix(acp)` | stop naming `--model` for a model that came from config |
| `fix(acp)` | keep session-creation model selection inside the acp layer |

The nine `fix` commits address defects in the feature commits above them, found in review before and during this PR. They are split out so each can be taken or dropped independently:

- A whitespace-only pin passed `z.string().min(1)`, won the precedence chain, then got trimmed away downstream — suppressing both fallbacks and requesting no model.
- `flow.json` described a model-pinned node identically to an unpinned one, so run bundles lost a field that changed how the run executed.
- The configured model was passed as the requested model on *every* prompt, and caller options beat the session record — so `set model` was reverted on the next turn, and `sessions ensure` reset an existing session the same way. It now travels as a creation-only `defaultModel`, resolved by `sessionCreationModel()` at each creation site; the prompt path still honors `model` alone, so `--model` continues to apply to a live session.
- The first cut of that fix resolved the default on helpers shared by the create **and** resume paths, so `session/resume` and `session/load` still re-sent it and took back a `set model` choice. The default is now resolved only when the ACP session is actually created; a resumed session honors an explicit `model` alone. (Reported by ClawSweeper; both call sites confirmed and fixed.)
- `acp` nodes share one session by default and a model is applied at creation, so a later node's pin was silently dropped — the `docs/flows.md` example was itself an instance of this. Conflicting pins on one session are now rejected instead of silently ignored. See the comment below for the contract and why. (Reported by ClawSweeper.)
- **Flows bypassed the creation-only channel entirely.** `mergeSessionModel` collapsed the node pin, the agent model, and `--model` into one explicit `model`, which is persisted with the session record, reloaded by `sendSessionDirect`, and re-requested by `applyPromptModelIfAdvertised` after reconnecting — so a node sharing a session re-sent the configured default on every later prompt. The agent model now travels as `defaultModel` here too; a node's own pin stays explicit, since it is a request from that node and conflicting pins are already rejected. The session binding records `sessionCreationModel(...)`, so the shared-session pin guard still sees a configured model. (Reported by ClawSweeper.)
- **That creation-only channel then had a hole of its own.** `assignClaudeCodeOptions` read `sessionOptions.model` only, so once the configured model became `defaultModel` it stopped reaching `_meta.claudeCode.options.model` on `session/new`. For a Claude ACP adapter that advertises no model list that is the *only* remaining channel — `assertRequestedModelSupported` deliberately lets such an adapter through without a config option and `applyRequestedModelIfAdvertised` then returns early — so the configured model was dropped silently, exit 0 and no warning, while `--model` worked. Reading it through `sessionCreationModel` restores the documented metadata channel without weakening the contract: `buildClaudeCodeOptionsMeta` has exactly one caller, `connection.newSession`, so it cannot reach an existing session.
- The unsupported-model errors said `Cannot apply --model "x"` no matter where the selector came from. With `agents.<name>.model` and node pins it can now arrive with no flag in sight, so the flag name is dropped.
- Reaching `sessionCreationModel` from those two call sites broke the declared dependency boundaries, since neither `src/acp` nor `src/flows` may depend on `src/runtime` (`ts.dependency-boundaries-required`). It moves to a leaf module under `src/acp`, which every layer that creates a session may reach, re-exported from its old home so existing callers are untouched, and typed structurally so the acp layer need not name `SessionAgentOptions`. `defaultModel` is also declared on `AcpClientOptions.sessionOptions`, which carried it at runtime while the type omitted it.

With no configured model and no node pin, behavior is unchanged: `defaultModel` is undefined and `sessionCreationModel()` returns exactly what `sessionOptions?.model` returned before.

Rebased onto current `main` (`035062a`); the only conflict was the `CHANGELOG.md` Unreleased list, resolved by keeping the README entry alongside both of this PR's.

## Testing

Fully tested — automated suite plus a live ACP adapter run (see below).

- `pnpm run check` green: format, typecheck, lint, build, viewer build, 913 + 128 tests.
- New unit coverage for config parsing, `resolveAgentInvocation`, CLI precedence and `sessionCreationModel`, flow precedence via `mergeSessionModel`, the whitespace rejection, the snapshot field, and the queue-owner IPC round-trip of `defaultModel`.
- New integration test asserting a configured model is requested at session creation via `session/set_config_option`, using the mock agent with `--advertise-models`.
- New coverage for the shared-session pin rule: conflict rejected, mutually exclusive `decision` branches accepted, unreachable nodes ignored, agent-level conflicts caught once agents resolve, and the runtime backstop across pinned/unpinned/legacy bindings.
- New regression tests for the resume paths: `runtime-manager` asserts the default applies on create but not on resume and that an explicit `--model` still applies on resume; `integration` asserts the same two through the CLI. Both were confirmed failing against the unfixed code.
- New regression tests for the three latest fixes, each confirmed failing first: a two-node flow sharing one session re-requested the configured model twice before and zero times after; a Claude-ACP-shaped adapter now carries the configured model in `session/new` metadata; the error message no longer names `--model`. A fourth test guards the flow *creation* path, which previously had no assertion that a configured model is still applied — it was proved to bite by temporarily breaking `mergeSessionModel` (`actual: 'default-model'`).

### Real-adapter proof

Adapter: `@agentclientprotocol/claude-agent-acp` driving Claude Code 2.1.220. Isolated `HOME`, config `agents.claude.model = "haiku"`, a scratch workspace.

Creation seeds from config, and `set model` then wins on the live session:

```text
$ acpx claude sessions new
[acpx] created session cwd (837a7cda-…)
  session record: current_model_id = haiku

$ acpx claude set model sonnet
model set: sonnet
  session record: current_model_id = sonnet
```

Resume, A/B against the same real session (whose model at the adapter is `opus[1m]`), same config, same command — only the build differs:

```text
$ acpx claude sessions new --resume-session 86138fb1-…

  A  pre-fix build (2322559)   -> current_model_id = haiku      # configured default clobbered the session
  B  fixed build   (89c2815)   -> current_model_id = opus[1m]   # session's own model preserved

$ acpx --model haiku claude sessions new --resume-session 86138fb1-…

  C  fixed build   (89c2815)   -> current_model_id = haiku      # an explicit --model still applies
```

A is the regression ClawSweeper flagged, reproduced against a real adapter; B and C are the fixed behavior.

The flow half of the feature, same adapter. Config `agents.claude.model = "opus[1m]"`, run-wide `--model haiku`, two `acp` nodes in isolated sessions — one pinned to `sonnet`, one unpinned:

```ts
pinned:   acp({ model: "sonnet", session: { isolated: true }, prompt: … }),
unpinned: acp({                  session: { isolated: true }, prompt: … }),
```

```bash
acpx --approve-all --model haiku flow run ./proof.flow.ts --default-agent claude
```

What the run bundle recorded, from the adapter's own reported session settings:

```text
flow.json  nodes.pinned.model   = "sonnet"     # node pin survives into the definition snapshot
flow.json  nodes.unpinned.model = absent

sessions/isolated-pinned-1-…    model = sonnet     # 1. node pin wins
sessions/isolated-unpinned-1-…  model = opus[1m]   # 2. agent config beats the run-wide --model haiku
```

That is the documented flow precedence — node → agent → `--model` — including the deliberate inversion versus the plain CLI, plus the `flow.json` snapshot field from the `fix(flows)` commit.

### Proof for the three latest fixes

These were verified against the mock agent and by tracing the creation path, not by a fresh live-adapter run; the live proof above still covers the feature itself.

The flow reconnect defect, from the run bundle's own ACP event log for the shared session:

```text
two acp nodes, one shared "main" session, agents.<name>.model = smart-model

  before  session/set_config_option(model) x2   # re-sent after reconnect
  after   session/set_config_option(model) x0   # creation only
```

The metadata hole, on a Claude-ACP-shaped adapter advertising no model list — same agent, same model, only the source differs:

```text
agents.claude.model = fast-model   _meta.claudeCode.options = { settingSources }              # dropped
--model fast-model                 _meta.claudeCode.options = { settingSources, model }
```

## AI assistance

AI-assisted (Claude Code). The feature commits were written with AI assistance; the review fixes came from this repo's `autoreview` helper (Codex) and from Claude Code review of the branch, verified against the code by hand before applying. I understand what the code does.


skywills:
Thanks — the shared-handle finding was correct, and it was the most useful one so far. Fixed in `410b89d`, along with the earlier resume P1s.

## The defect

`ensureSessionBinding` keys on `agentCommand::cwd::handle` and the default handle is `main`, so **every `acp` node shares one session unless it opts out**. The reuse branch returns before any model is considered, so a later node's pin was silently dropped.

The example I wrote in `docs/flows.md` was itself an instance of it:

```ts
summarize: acp({ model: "gpt-5-mini", ... }),   // creates the shared "main" session
fix:       acp({ model: "gpt-5-codex", ... }),  // silently runs on gpt-5-mini
```

Which is why the flow proof in the PR body needed `session: { isolated: true }` on both nodes to show anything at all. You get billed for the wrong model with no warning, and which model you get depends on execution order.

## The contract

I took your recommended option — **pins are session-bound**. A model belongs to a session, not to a step, so a node that cannot have its pin honored is an error rather than a silent downgrade:

- Different pins on one session → **rejected**, with the error naming both nodes and pointing at `isolated: true` or a distinct handle.
- An **unpinned** node sharing a session still inherits — it asked for no particular model, so there is nothing to violate.
- The run-wide `--model` is deliberately **not** a pin. It is a default for every step, so a node that merely falls through to it inherits rather than conflicting.
- A configured `agents.<name>.model` **is** a pin, since flow precedence already ranks it above `--model`. Two profiles resolving to the same command, cwd, and handle therefore conflict too.

I did not take the "apply the pin per step" reading. It would make the docs example work as written, but it mutates shared session state — an unpinned node running after a pinned one would inherit the pin — and it cuts against the creation-only contract the CLI side of this PR establishes.

## Enforcement

Three layers, so the common case fails before spending anything and nothing silently slips through:

| Layer | Sees | Catches |
| --- | --- | --- |
| `validateFlowDefinition` | node pins | conflicts visible in the flow file alone |
| runner pre-flight | + resolved agents | agent-model conflicts, before any step runs |
| `assertReusableSessionModel` | the actual binding | computed `cwd`, anything genuinely dynamic |

Static analysis only compares nodes that can actually collide: isolated nodes are excluded, unreachable nodes are ignored, and two nodes must be reachable from one another to be compared — so a `decision` routing to differently pinned branches stays valid, since only one branch ever creates the session.

One compatibility note, since flow node `model` currently passes `.passthrough()` on `main` and is ignored: run bundles exist whose session bindings have no `model` field. Treating that absence as "created without a model" would break resuming an in-progress run after upgrade, so a binding now records `null` for "created with no model" and absent still means "predates this field", which skips the guard.

## Verified

Against a live `@agentclientprotocol/claude-agent-acp` adapter, config `agents.claude.model = "opus[1m]"`, run-wide `--model haiku`:

```text
valid flow, one pinned + one unpinned isolated node   exit 0
  isolated-pinned-…    model = sonnet     # node pin wins
  isolated-unpinned-…  model = opus[1m]   # agent config beats --model haiku

conflicting flow, two pins on the shared "main" session   exit 1
  Flow nodes "cheap" and "heavy" share session handle "main" but pin
  different models ("haiku" vs "sonnet"). …
  run bundles created: 0 — rejected before any step ran
```

`pnpm run check` is green: 906 + 125 tests.

## Still yours to decide

Whether `agents.<name>.model` and node `model` should become stable public configuration surface with intentionally different CLI and flow precedence, per your Decision needed. This picks one coherent reading and documents it, but the product call is yours — happy to change the semantics or narrow the surface if you would rather it landed differently.

skywills:
Reconnect finding fixed in `73c1fc4`. Rebased onto `035062a`, head `2e45cbf`, conflict resolved.

**The finding.** `mergeSessionModel` collapsed the node pin, `agents.<name>.model`, and the run-wide `--model` into one explicit `model` — persisted with the record, reloaded by `sendSessionDirect`, re-requested by `applyPromptModelIfAdvertised` after reconnecting. The agent model now travels as the creation-only `defaultModel`, which is never persisted; a node's own pin stays explicit, and the binding records `sessionCreationModel(...)` so the shared-session guard still sees a configured model. From the run bundle's ACP event log, two nodes sharing one session: **2** `session/set_config_option(model)` before, **0** after.

**Two defects it exposed.**

- `assignClaudeCodeOptions` read `sessionOptions.model` only, so once the configured model became `defaultModel` it stopped reaching `_meta.claudeCode.options.model` on `session/new`. For a Claude ACP adapter advertising no model list that is the only channel left, so the configured model was dropped silently — exit 0, no warning — while `--model` worked. This was live on the CLI path too, from `877b377`; not introduced by the flows change. Fixed in `0873e9a`.
- Unsupported-model errors named `--model` regardless of where the selector came from (`9269324`), and the shared helper had to move out of `src/runtime` to satisfy `ts.dependency-boundaries-required` (`2e45cbf`).

`pnpm run check` green: 913 + 128 tests, CI green. Each fix has a regression test confirmed failing against the unfixed code; the flow-creation guard was proved to bite by temporarily breaking `mergeSessionModel`.

Proof for these three is mock-agent and creation-path tracing, not a fresh live-adapter run — the live proof in the PR body still covers the feature itself.

Your **Decision needed** is unchanged and still yours: whether `agents.<name>.model` and node `model` should become stable public configuration surface with deliberately inverted CLI and flow precedence.

@clawsweeper re-review

skywills:
The pass-through compatibility issue is partially addressed as of `358946b`: a legacy `model` value that's non-string or blank no longer throws on upgrade (previously any non-empty-string-required validation would reject a formerly valid flow).

What remains open, per ClawSweeper's latest review: a pre-existing flow node where `model` held an unrelated non-blank string will still be picked up as a real model pin after upgrade, since nothing distinguishes "old inert string" from "new intentional pin" once both are just non-blank strings under the same key. Closing this fully requires reserving a distinct field name for the pin (e.g. `modelPin`) and leaving `model` permanently inert — a larger change touching `types.ts`, `schema.ts`, `model-pins.ts`, `runtime.ts`, `store.ts`, docs, and tests.

This is the owner decision the review keeps surfacing: should the pin live under a new key (full backward compatibility, more surface area, delayed ship) or is reserving `model` an acceptable documented breaking change (ships as-is, existing flows using `model` for something else must rename on upgrade)? Happy to implement either once there's a call — didn't want to do the rename speculatively before knowing which way you want it.

steipete:
Closing after review: this reuses formerly arbitrary pass-through `model` metadata in a way that isn't safe to trust, misses the Claude guide synchronization, and spreads a contract change across 33 files. The per-agent/per-node model idea itself has merit — a scoped follow-up that introduces a validated model field (not reusing the legacy pass-through) would be reviewable. Thanks @skywills.

PR #491, opened 2026-08-05
fix(session): re-apply pinned model before set on reconnect

### LINKED ISSUE #489: set <config-option> fails after reconnect because the pinned model is not re-applied on a resumed session
## Summary

When a session is pinned to a non-default model at creation (`--model`), any later
`acpx <agent> set <configId> <value>` runs against a session that has silently
reverted to the agent's *default* model. For agents whose config options are
model-dependent (opencode's `effort`), the set is rejected by the agent because
the advertised option set belongs to the wrong model.

The prompt path is unaffected — it re-applies the pinned model after connecting.
Only the `set` paths are broken.

## Environment

- acpx 0.13.0 (upstream/main, 035062a)
- opencode-ai 1.18.13
- Linux, Node via npx

## Reproduction

```bash
acpx --model opencode-go/deepseek-v4-flash opencode sessions ensure --name "acpx-test-model"
acpx opencode set effort high -s "acpx-test-model"
```

Expected: effort is set to `high`.

Actual:

```
Agent rejected session/set_config_option for "effort"="high":
Invalid params: effort not found: high (ACP -32602).
```

Passing `--model opencode-go/deepseek-v4-flash` on the `set` command does not help.

## Observed session-record state

Tracking `acpx.current_model_id` and the advertised `effort` options in
`~/.acpx/sessions/<id>.json` across the three commands:

| step                                              | current_model_id                | effort options   |
|---------------------------------------------------|---------------------------------|------------------|
| after `sessions ensure --model deepseek-v4-flash`  | `opencode-go/deepseek-v4-flash` | `low, high, max` |
| after `set effort high` (new process)              | `minimax/MiniMax-M3`            | `none, thinking` |
| after any `prompt`                                 | `opencode-go/deepseek-v4-flash` | `low, high, max` |

`acpx.session_options.model` stays `opencode-go/deepseek-v4-flash` throughout, so
acpx does know the pin — it just doesn't apply it.

## Root cause

Each acpx invocation is a separate process, so the agent is respawned and the ACP
session is resumed/loaded. opencode does not restore the per-session model on
resume; it comes back on the configured default, and its `effort` option set is
model-dependent.

`replayFreshSessionPreferences` (src/runtime/engine/reconnect.ts:461) skips all
preference replay unless the session had to be created fresh:

```ts
if (!params.createdFreshSession) {
  return { modelReplay: { replayed: false }, configReplay: { replayed: false } };
}
```

The prompt path compensates independently: it merges
`sessionOptionsFromRecord(record)` (src/cli/session/runtime.ts:696) and calls
`applyPromptModelIfAdvertised` after connect (src/cli/session/runtime.ts:1021).

The set paths — `runSessionSetConfigOptionDirect`, `runSessionSetModeDirect`,
`runSessionSetModelDirect` (src/cli/session/prompt-runner.ts:124-185) — have no
equivalent step, so they issue `session/set_config_option` against a session that
has reverted to the default model.

## Secondary effect: the record is left advertising the wrong options

Even though the set fails, the resumed session's config options are still written
back to the record, so `current_model_id` and `config_options` are overwritten
with the *default* model's values. Subsequent `status` / option listings then
advertise options the pinned model does not have, until the next prompt re-pins
the model.

## Possible fixes

1. Narrow — mirror the prompt path in `prompt-runner.ts`: re-apply
   `sessionOptionsFromRecord(record).model` after connect, before the `set_*` call.
2. Broader — in `connectAndLoadSession`, replay the desired model on resumed
   sessions too when the agent reports a different `current_model_id`. This fixes
   `set`, `sessions ensure`, `status` and anything else on that path in one place.

(2) touches the intent documented at src/cli/session/session-management.ts:161 —
"a resumed ACP session already exists, so only an explicit `model` may change it;
a configured default seeds creation alone" — so it needs a maintainer call on
whether a pin stored in `session_options.model` counts as explicit on resume.
Note the prompt path already behaves as (2) today, so the two paths are
currently inconsistent.

SebTardif:
## What Problem This Solves

Fixes an issue where users who pin a non-default model at session creation (`--model`) then run `acpx <agent> set <configId> <value>` against that session would hit agent rejection for model-dependent options (for example opencode `effort`). Each set is a new process that resumes the ACP session; some agents restore their default model on resume, and the set paths never re-applied the pin. The prompt path already did.

## Why This Change Was Made

After reconnect, the set paths (`set_mode`, `set_model`, `set_config_option`) now re-apply `session_options.model` before mutating config, matching the prompt path. Scope is intentionally narrow: only the set entrypoints, not a full reconnect-preference policy change.

## User Impact

Pinned-model sessions keep working for config sets after process restart. Operators no longer need to re-prompt solely to restore the model before a set.

## Evidence

Production path: `withConnectedSession` → set direct helpers in `prompt-runner.ts`. After this patch, a resumed set re-pins `session_options.model` before `session/set_config_option`.

Red/green on a load-capable agent that starts each process at the default model while the session record keeps `session_options.model = smart-model`:

```console
$ node --test dist-test/test/prompt-runner.test.js
(without fix) re-applies pinned model after reconnect FAIL
  Expected current_model_id smart-model, got default-model

(with fix) re-applies pinned model after reconnect PASS
  current_model_id = smart-model
  model option currentValue = smart-model
  reasoning_effort = high
```

## Real behavior proof

- **Behavior or issue addressed:** Re-apply the session-pinned model on set after reconnect so model-dependent config options are advertised for the pin, not the agent default.
- **Real environment tested:** macOS arm64, Node 26, worktree `/tmp/acpx-489` on `fix/489-reapply-pinned-model-on-set` from `upstream/main`.
- **Exact steps or command run after this patch:** `pnpm run build:test` then `node --test dist-test/test/prompt-runner.test.js` (red without the re-apply helper, green with it).
- **Evidence after fix:** console output above; after reconnect, set reasoning_effort high leaves current_model_id and model option at smart-model.
- **Observed result after fix:** pinned model survives set; desired_config_options and advertised options stay on the pin.
- **What was not tested:** live opencode binary with deepseek-v4-flash effort options (issue repro uses that agent; this path uses the in-repo load-capable agent that resets model on new process, matching the same failure mode).

## Summary

- Add `reapplyPinnedModelAfterConnect` before set_mode / set_model / set_config_option
- Regression covering reconnect + set config with session_options.model pin
- Changelog under Unreleased Fixes

## Testing

- Red: new case failed without the fix (current_model_id fell back to default-model)
- Green: `node --test dist-test/test/prompt-runner.test.js` → 5 passed
- oxlint/oxfmt clean on changed files; typecheck via build:test

Closes #489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Repo: https://github.com/openclaw/acpx
 ### Fixes
 
 - Runtime/embedding: settle turn results only after lifecycle persistence and client cleanup attempts finish, including unexpected finalization failures.
+- CLI/session: re-apply a session-pinned model before set_mode/set_model/set_config_option after reconnect, matching the prompt path so model-dependent options work. Fixes #489.
 
 ## 2026.7.27 (v0.13.0)
 
--- a/src/cli/session/prompt-runner.ts
+++ b/src/cli/session/prompt-runner.ts
@@ -198,14 +198,8 @@ export async function runSessionSetModelDirect(
 ): Promise<SessionSetModelResult> {
   const result = await withConnectedSession(
     buildDirectConnectedSessionOptions(options, async ({ client, sessionId, record }) => {
-      // set-model is an explicit model change; still re-apply pin first so the
-      // agent advertises the pinned option set before we switch.
-      await reapplyPinnedModelAfterConnect({
-        client,
-        sessionId,
-        record,
-        timeoutMs: options.timeoutMs,
-      });
+      // Explicit model switch replaces the saved pin. Do not re-apply the old
+      // pin first: an unadvertised saved model would reject and block the switch.
       const models = advertisedModelState(record.acpx);
       const response = await withTimeout(
         client.setSessionModel(sessionId, options.modelId, models),
@@ -226,13 +220,18 @@ export async function runSessionSetConfigOptionDirect(
 ): Promise<SessionSetConfigOptionResult> {
   const result = await withConnectedSession(
     buildDirectConnectedSessionOptions(options, async ({ client, sessionId, record }) => {
-      await reapplyPinnedModelAfterConnect({
-        client,
-        sessionId,
-        record,
-        timeoutMs: options.timeoutMs,
-      });
       const modelConfigId = advertisedModelState(record.acpx)?.configId;
+      // Model-valued config updates replace the pin; skip pin replay so an
+      // obsolete saved model cannot block the replacement. Other config keys
+      // still re-apply the pin so model-dependent options see the right set.
+      if (options.configId !== modelConfigId) {
+        await reapplyPinnedModelAfterConnect({
+          client,
+          sessionId,
+          record,
+          timeoutMs: options.timeoutMs,
+        });
+      }
       const response = await withTimeout(
         client.setSessionConfigOption(sessionId, options.configId, options.value),
         options.timeoutMs,
--- a/src/cli/session/prompt-runner.ts
+++ b/src/cli/session/prompt-runner.ts
@@ -1,3 +1,4 @@
+import type { AcpClient } from "../../acp/client.js";
 import { withTimeout } from "../../async-control.js";
 import {
   withConnectedSession,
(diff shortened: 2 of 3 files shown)


SebTardif:
@clawsweeper re-review

Addressed the latest durable review findings on this head. Please re-evaluate.

SebTardif:
@clawsweeper re-review

Please re-evaluate on the current head. Prior durable review findings should already be addressed in code.

PR #498, opened 2026-08-13
fix(session): completed turn stays successful when session save fails

SebTardif:
Fixes an issue where a persistent ACP turn could finish as successful after the final session write failed.

Related: https://github.com/openclaw/acpx/pull/495

## What Problem This Solves

Fixes an issue where users running a persistent ACP turn would see the turn complete successfully even when the final session checkpoint flush or session-store save failed (for example, disk full). The next restart then resumed from stale or missing state.

## Why This Change Was Made

https://github.com/openclaw/acpx/pull/495 already fails the turn when unexpected finalization work throws. The last checkpoint flush and session-store save still used empty `.catch(() => {})`, so persist failures never reached that path. This change lets those errors propagate. Streamed ACP events remain available; `result` now matches whether the session was actually saved.

## User Impact

`startTurn()` and `runTurn()` callers get `status: failed` when the post-turn persist step cannot write the session. The client is not left pooled as if the save succeeded. CLI process-exit best-effort flush is unchanged.

## Evidence

terminal output from production `createAcpRuntime` + `createFileSessionStore` against a real mock-agent child. First turn writes `~state/sessions/proof-498-persist.json`. On the next turn, the finalize persist step is injected with `ENOSPC`.

Unpatched `upstream/main` `1402be3` still reports that turn completed:

```console
$ node /tmp/proof-498-persist.mjs /tmp/acpx-498-main/dist/runtime.js dist-test/test/mock-agent.js main
{
  "label": "main",
  "first": { "status": "completed", "stopReason": "end_turn", "sessionBytes": 1743 },
  "second": {
    "status": "completed",
    "stopReason": "end_turn",
    "error": null,
    "text": "unrecognized prompt: second persist",
    "savesAfterPrompt": 3
  }
}
```

Patched `4bd1bf5` (same tree as `2967601`) fails the same finalize persist and still delivers the agent text first:

```console
$ node /tmp/proof-498-persist.mjs /tmp/acpx-498/dist/runtime.js dist-test/test/mock-agent.js patched
{
  "label": "patched",
  "first": { "status": "completed", "stopReason": "end_turn", "sessionBytes": 1746 },
  "second": {
    "status": "failed",
    "error": { "message": "ENOSPC: no space left on device", "code": "RUNTIME" },
    "text": "unrecognized prompt: second persist",
    "savesAfterPrompt": 3
  }
}
```

## Real behavior proof

[...]

PR #501, opened 2026-08-16
fix(acp): time out hung process-list helpers after terminal exit

SebTardif:
## What Problem This Solves

Fixes an issue where users waiting on an ACP shell-backed terminal would hang forever in `terminal/wait_for_exit`, `terminal/kill`, or `terminal/release` after the command had already exited, when the follow-up `ps` or PowerShell process-list helper never returned.

The same unbounded helper class could also stall session close (matching a leftover owner via `ps` / PowerShell) and WSL cwd translation (`wslpath`).

## Why This Change Was Made

`exitPromise` waits on `rememberProcessGroupPids()`, which used to spawn `ps` / PowerShell with no timeout. A stuck helper kept wait, kill, and release parked even though the terminal process itself was gone.

This change runs those short helpers through a shared 8s `execFile` timeout (SIGKILL on expiry) and treats a timeout as "no process list" so release can continue. Session-close process matching and `wslpath` use the same helper. Agent `initialize` is not given a generic timeout; only Gemini remains timed by design.

## User Impact

A finished terminal can be waited on, killed, or released even when the host process listing tool is stuck. Session close and WSL cwd translation fail closed or skip the match instead of hanging the CLI.

## Evidence

Live `node` against production `TerminalManager` and `runTimedExecFile` from the patched worktree at `/tmp/oc-impl-acpx-ps-timeout`. A fake `ps` that only sleeps was placed first on `PATH`. Same script compared the old unbounded spawn with the patched wait/release path.

```console
$ node /tmp/acpx-ps-timeout-proof.mjs
{
  "beforeUnboundedPs": {
    "result": "hung",
    "elapsedMs": 1502
  },
  "timedHelper": {
    "result": "timed_out",
    "elapsedMs": 506,
    "killed": true,
    "signal": "SIGKILL"
  },
  "afterTerminalManager": {
    "waitMs": 506,
    "releaseMs": 1007,
    "exitCode": 0,
    "signal": null
  }
}
```

The unpatched `ps` spawn was still running at 1502ms and had to be killed by the proof script. After the patch, `terminal/wait_for_exit` returned `exitCode: 0` in 506ms and `terminal/release` finished in 1007ms while the same hanging `ps` stayed on `PATH`.

Introduced in [#313](https://github.com/openclaw/acpx/pull/313) (2026-05-15) when shell-backed terminals started snapshotting the process group after exit. Not [#498](https://github.com/openclaw/acpx/pull/498) (persist swallow).

## Real behavior proof

- **Behavior or issue addressed:** After a shell-backed ACP terminal exits, wait_for_exit, kill, and release hang if the `ps` / PowerShell process-list helper never returns.
- **Real environment tested:** macOS 26.6.1, Node v26.7.0, patched acpx checkout at `/tmp/oc-impl-acpx-ps-timeout`.
- **Exact steps or command run after this patch:**

  ```bash
  cd /tmp/oc-impl-acpx-ps-timeout
  node /tmp/acpx-ps-timeout-proof.mjs
  ```

- **Evidence after fix:** terminal output from the live node script above. Unbounded `ps` stayed hung at 1502ms. `runTimedExecFile` killed the helper at 506ms (`killed: true`, `SIGKILL`). Production `TerminalManager.waitForTerminalExit` then returned `exitCode: 0` in 506ms and `releaseTerminal` completed in 1007ms with the hanging `ps` still first on `PATH`.
- **Observed result after fix:** wait and release no longer wait on the stuck process-list helper. The terminal exit code is still reported. The hung helper is SIGKILL'd instead of left running.
- **What was not tested:** Windows PowerShell listing under a live hung `powershell.exe`, and a real WSL `wslpath` hang. Those call the same timed helper.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Repo: https://github.com/openclaw/acpx
 - Runtime/embedding: snapshot permission policies at client configuration boundaries so caller-side mutation cannot change an in-flight turn after prompt readiness.
 - Replay viewer: return the same not-found response for missing files and containment-denied paths, preventing outside-target existence probes.
 - Runtime/sessions: surface checkpoint flush and session-store save failures during turn finalization instead of dropping them. Thanks @SebTardif.
+- ACP/terminal: time out hung `ps` / PowerShell process-list helpers after a shell-backed terminal exits so wait_for_exit, kill, and release can finish. Thanks @SebTardif.
 
 ## 2026.7.27 (v0.13.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Repo: https://github.com/openclaw/acpx
 - Replay viewer: return the same not-found response for missing files and containment-denied paths, preventing outside-target existence probes.
 - Runtime/sessions: surface checkpoint flush and session-store save failures during turn finalization instead of dropping them. Thanks @SebTardif.
 - ACP/terminal: time out hung `ps` / PowerShell process-list helpers after a shell-backed terminal exits so wait_for_exit, kill, and release can finish. Thanks @SebTardif.
+- ACP/terminal: raise the process-list helper stdout cap above execFile's 1 MiB default so large `ps` listings are not dropped as empty. Thanks @SebTardif.
 
 ## 2026.7.27 (v0.13.0)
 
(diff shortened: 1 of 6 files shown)


SebTardif:
@clawsweeper

> [P2] Preserve process-list output beyond execFile's default buffer
> execFile defaults to a 1 MiB stdout buffer, while the replaced spawn readers streamed the complete system process list. On a host whose ps -eo output crosses that threshold this helper rejects, both list functions silently return [], and post-exit descendant cleanup can miss live processes.

Pushed [`d6b7581`](https://github.com/openclaw/acpx/commit/d6b7581). The timed helper now sets maxBuffer to 32 MiB.

@clawsweeper re-review

PR #522, opened 2026-08-27
fix(runtime): settle session work before fresh preparation

vincentkoc:
Fresh-session preparation could acknowledge a reset while an earlier turn was still writing the same record. That late write could erase the reset or restore the old backend identity. A failed cleanup could also leave an unusable owner available for reuse, or be forgotten when preparation was retried.

The public `prepareFreshSession({ handle })` API already landed in https://github.com/openclaw/acpx/pull/596. This PR now focuses on completing its lifecycle guarantee:

- Admit complete turns and preparation in order for each exact session record. Different records remain concurrent, and controls still reach the active turn.
- Cancel and settle previously submitted turns before publishing fresh-session intent. A queued cancellation never starts its prompt or cancels its predecessor.
- Unpublish owners before destructive cleanup, retain failed cleanup for retry, and coalesce concurrent cleanup attempts. Preparation rejects while the original owner still cannot be finalized.
- Cancel queued work on ordinary close while preserving its existing early-return and resumable-session behavior. Remote discard retains its protocol requirement; fresh preparation does not require remote close or delete native history.

The original contribution's completion and strict-finalization work informed this repair. It preserves the accepted object-shaped API and uses the valid identical-launch restart control already on main, replacing the earlier process fixture that could pass solely because launch identity changed. Thanks @vincentkoc.

### Reproductions

The regression cases failed before their fixes for the intended reasons:

```text
active preparation: acknowledged before the active turn finalized
multiple submissions: only 1 of 2 turns finalized before preparation returned
failed idle cleanup: retry lost its session update
cleanup retry: preparation falsely succeeded while owner finalization still failed
ordinary close: a queued prompt still ran (2 prompts instead of 1)
```

### Validation

- Full `pnpm run check` and `pnpm run check:docs` passed on an isolated Linux runner.
- All 117 focused runtime, manager, and session-context tests passed; regression cases include concurrent records, responsive active controls, queued cancellation, failed cleanup retries, and restart identity.
- Independent structured review: no actionable P0–P2 findings (confidence 0.96).
- Source-blind public-runtime proof passed against real stdio ACP subprocesses and file storage. Active plus queued preparation returned only after both turns were cancelled; queued text was suppressed; fresh identity remained stored. Ordinary close retained its permitted early return, both prior turns later cancelled, and the record remained closed.
- Real Codex adapter smoke passed on the final build: preparation returned after `cancelled/cancelled` in 1,268 ms, then a runtime restart produced a different backend under the same acpx record and delivered fresh synthetic text. A separate real native-history control successfully resumed the retired backend and recalled its synthetic marker; that unchanged-history proof was reused in the final focused rerun.
- All live tests used synthetic state. Owned runtimes and child processes were stopped; synthetic state was cleaned up.

This repair changes production by +70 net lines, tests by +210, and docs/changelog by +3. The added admission and retry state is confined to the existing runtime manager; the submitted tests are substantially smaller than the original PR's process harness.


steipete:
Maintainer triage: this remains a product/API decision because it introduces public `prepareFreshSession(handle)` semantics. There is also a concrete problem with the supplied restart proof.

`test/runtime-retirement-process.ts` seeds a record with `agentCommand: "fixture"` and no `agentArgv`, while the ensure process resolves `fixture` to `[process.execPath, path.resolve(mockAgentPath)]`. The existing reuse policy rejects a command/argv mismatch independently of retirement. Consequently, the three-process fixture can allocate a fresh backend even if the retirement phase is omitted; a changed backend UUID alone does not prove the new marker works.

The meaningful control is identical launch identity in every process: create the original record through real ensure, prove restart without retirement keeps the backend identity, then prove retirement changes that identity while preserving the intended record identity. This is a source-supported assessment, not a claim that I executed the retirement scenario.

Recommendation to the orchestrator: keep the decision with Peter until the retirement API, treatment of existing handles, and persistence guarantee are selected; the current proof should not be treated as satisfying that guarantee. No implementation or contributor changes are requested by this triage pass.

steipete:
Maintainer update: https://github.com/openclaw/acpx/pull/596 already established the object-shaped fresh-session API. This rewrite keeps that accepted public contract and completes its cancellation, finalization, and cleanup-retry guarantees in the existing runtime manager. Same-record work is serialized; different records remain concurrent. Ordinary close retains its existing early-return/resume behavior.

The former API-absence blocker is superseded. The earlier false-positive restart fixture is replaced by the identical-launch control already on main, plus active/queued-turn and cleanup-failure regressions. The updated body includes observed failures before the fix, passing isolated Linux checks, independent review, and source-blind real Codex cancellation/restart/native-history evidence. Thanks @vincentkoc for the original retirement and finalization work.

steipete:
Landed as 53177fc9f04ce913497f75c3fc9793382a24ae58 from reviewed head 707a72b42a9d6e47ead09b93f45acfe31899f453.

Validation: 117 focused runtime/manager/session-context tests; full isolated Linux `pnpm run check` and `pnpm run check:docs`; independent review with no actionable findings; and green [exact-head CI](https://github.com/openclaw/acpx/actions/runs/35077641510), including Node 22/24/26 and mutation tests. ClawSweeper's refreshed review also found no blockers.

Source-blind proof exercised real stdio/file-store cancellation, queued work, cleanup failure/retry, and restart. Real Codex preparation returned after explicit cancellation, and restart produced a new backend under the same acpx record with successful text output. Separate native-history proof reopened the retired backend and recalled its synthetic marker. All task-owned test processes and state were cleaned up; main is synchronized and clean.

Thanks @vincentkoc for the original contribution.

PR #529, opened 2026-08-29
fix(flows): ignore EPIPE on shell action stdin

SebTardif:
## What Problem This Solves

Fixes an issue where users running `defineFlow` / `shell()` / `acpx flow` with `ShellActionExecution.stdin` would lose the whole CLI when the child exited without reading that stdin. Node treats an `error` event on the child stdin pipe as fatal when no listener is attached, so a large write (1 MiB) to a child that does `setImmediate(() => process.exit(0))` raised uncaught `EPIPE` and killed the host.

## Why This Change Was Made

`runShellAction` already pipes stdin and settles the action from the child `exit` event. It did not attach an `error` listener on `child.stdin`, and it called `write` / `end` even after the stream had already ended. This change ignores pipe-death codes on stdin (`EPIPE`, `EIO`, `ECONNRESET`, `ERR_STREAM_DESTROYED`) and skips `write` / `end` when the stream is no longer writable. The existing `exit` handler still records status.

This is not the CLI `process.stdout` path from closed [#441](https://github.com/openclaw/acpx/pull/441). That helper can throw or exit. Child stdin must not. Open [#507](https://github.com/openclaw/acpx/pull/507) covers terminal stdout/stderr listeners. Merged [#501](https://github.com/openclaw/acpx/pull/501) timed out hung process-list helpers. Merged [#498](https://github.com/openclaw/acpx/pull/498) covers finalize persist errors. None of those touch flow shell stdin.

The missing listener dates to [#179](https://github.com/openclaw/acpx/pull/179) (`697ee1f`, 2026-03-26, "feat: add experimental acpx flows runtime and examples"), 156 days on `main`.

## User Impact

A flow that feeds stdin into a short-lived shell child keeps the ACP host alive. The shell action still completes from the child's exit status. Operators no longer lose the CLI because leftover stdin raised `EPIPE`.

## Evidence

terminal output from live `node` against compiled public `defineFlow` / `shell()` / `FlowRunner` (`dist/flows.js`).

Same script: 1 MiB `stdin` to `process.execPath -e "setImmediate(() => process.exit(0))"`.

Before, on unpatched `src/flows/executors/shell.ts` compiled to `dist`, the host throws and exits 1:

```console
$ node /tmp/acpx-F003-proof.mjs
node:events:505
    throw er; // Unhandled 'error' event
    ^

Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)
[...]

PR #530, opened 2026-08-29
fix(flows): kill shell children on interrupt and node timeout

SebTardif:
## What Problem This Solves

Fixes an issue where users running `acpx flow` / `FlowRunner.run` with a `shell()` node would leave the spawned command running after Ctrl+C or after an outer node timeout. `runShellAction` only sent SIGTERM/SIGKILL on its own `timeoutMs`. `FlowRunner.run` persisted the interrupted bundle and rejected; the CLI then `process.exit`ed and orphaned the child. The same orphan happens when the node timer fires (`timeoutMs: 200`) while the action sets `execution.timeoutMs: 0`, so the inner kill timer never starts.

## Why This Change Was Made

`runShellAction` now accepts an `AbortSignal`. Abort and the existing inner timeout share the same SIGTERM then SIGKILL grace (1s). `FlowRunner.run` aborts that signal on SIGINT/SIGTERM/SIGHUP. `executeActionNode` also aborts it from `runWithHeartbeat`'s `onTimeout`, matching how ACP nodes cancel a session on node timeout.

This is not stdin EPIPE handling (open [#529](https://github.com/openclaw/acpx/pull/529)). It is not output bounding. It does not change successful shell completion.

The missing abort dates to [#179](https://github.com/openclaw/acpx/pull/179) (`697ee1f`, 2026-03-26, "feat: add experimental acpx flows runtime and examples"), 156 days on `main`. [#188](https://github.com/openclaw/acpx/pull/188) persisted interrupted runs and still did not reap the spawn.

## User Impact

A flow `shell()` child is reaped when the operator interrupts the run or when the node timeout fires, including when the action sets `timeoutMs: 0`. The host still reports `TimeoutError` / `InterruptedError`. Operators no longer leak long-running `sleep` / compiler / `yes` processes after the CLI has exited.

## Evidence

terminal output from live `node` against compiled public `defineFlow` / `shell()` / `FlowRunner` (`dist/flows.js`).

Same scripts: one-node flow whose `exec` returns `command: process.execPath`, `args` that write `process.pid` then `setInterval`, and `timeoutMs: 0`. Outer node timeout is 250ms. Interrupt script starts the same flow in a child host, then sends SIGINT.

Before, on unpatched `src/flows/executors/shell.ts` and `src/flows/runtime.ts` compiled to `dist`, the host rejects but the child stays alive:

```console
$ ACPX_ROOT=/tmp/oc-pr-acpx-F004 node /tmp/acpx-F004-proof.mjs
{
  "hostAlive": true,
  "hostPid": 56463,
  "status": "rejected",
  "errorName": "TimeoutError",
  "errorMessage": "Timed out after 250ms",
  "childPid": 56471,
  "childAliveAfterTimeout": true
}
```

```console
$ SIGINT to FlowRunner host after the shell child wrote its pid
{
  "hostPid": 56490,
  "hostExit": 130,
  "childPid": 56499,
  "childAliveBeforeSigint": true,
  "childAliveAfterHostExit": true
}
```

After, on the patched compiled public API, the same flows reject and the child is gone:

```console
$ ACPX_ROOT=/tmp/oc-pr-acpx-F004 node /tmp/acpx-F004-proof.mjs
{
  "hostAlive": true,
  "hostPid": 56583,
  "status": "rejected",
  "errorName": "TimeoutError",
  "errorMessage": "Timed out after 250ms",
  "childPid": 56590,
  "childAliveAfterTimeout": false
}
```

```console
$ SIGINT to FlowRunner host after the shell child wrote its pid
{
  "hostPid": 56619,
  "hostExit": 130,
  "childPid": 56627,
  "childAliveBeforeSigint": true,
  "childAliveAfterHostExit": false
}
```

## Real behavior proof

- **Behavior or issue addressed:** Flow `shell()` children stayed running after Ctrl+C and after an outer node timeout when `execution.timeoutMs` was 0, so `process.exit` orphaned the spawn.
- **Real environment tested:** macOS, Node v26.7.0, acpx compiled from this branch at `/tmp/oc-pr-acpx-F004` (`dist/flows.js`).
- **Exact steps or command run after this patch:**

  ```bash
  ACPX_ROOT=/tmp/oc-pr-acpx-F004 node /tmp/acpx-F004-proof.mjs
  ACPX_ROOT=/tmp/oc-pr-acpx-F004 ACPX_F004_PID=/tmp/acpx-f004-int.pid node /tmp/acpx-F004-interrupt-host.mjs
  ```

  The timeout script imports `defineFlow`, `shell`, and `FlowRunner` from compiled `dist/flows.js`, then runs a one-node flow with node `timeoutMs: 250` and action `timeoutMs: 0`. The interrupt host runs the same long child; the parent sends SIGINT after the pid file appears.

- **Evidence after fix:** terminal output from the patched compiled public API (shown above). Timeout path prints `childAliveAfterTimeout: false`. Interrupt path prints `childAliveAfterHostExit: false` and host exit 130.
- **Observed result after fix:** The same pid-writing `setInterval` child that stayed alive after TimeoutError and after SIGINT on unpatched `dist` is now gone once the host rejects. The host still surfaces `TimeoutError` / `InterruptedError`.
- **What was not tested:** Windows cmd.exe spawn, and a child that ignores SIGTERM for the full 1s grace before SIGKILL when the CLI `process.exit`s immediately.

Related: [#179](https://github.com/openclaw/acpx/pull/179) (origin), [#188](https://github.com/openclaw/acpx/pull/188) (interrupt persist only), [#529](https://github.com/openclaw/acpx/pull/529) (stdin EPIPE, different file), [#501](https://github.com/openclaw/acpx/pull/501), [#498](https://github.com/openclaw/acpx/pull/498). Same leftover-child class as merged process-list timeouts in [#501](https://github.com/openclaw/acpx/pull/501).

### DIFF
--- a/src/flows/executors/shell.ts
+++ b/src/flows/executors/shell.ts
@@ -1,6 +1,15 @@
-import { spawn } from "node:child_process";
-import { TimeoutError } from "../../async-control.js";
+import { spawn, type ChildProcess, type ChildProcessWithoutNullStreams } from "node:child_process";
+import { InterruptedError, TimeoutError } from "../../async-control.js";
 import type { ShellActionExecution, ShellActionResult } from "../runtime.js";
+
+const SHELL_CHILD_KILL_GRACE_MS = 1_000;
+
+function terminateShellChild(child: ChildProcess): void {
+  child.kill("SIGTERM");
+  setTimeout(() => {
+    child.kill("SIGKILL");
+  }, SHELL_CHILD_KILL_GRACE_MS).unref();
+}
 
 export function formatShellActionSummary(spec: ShellActionExecution): string {
   return `shell: ${renderShellCommand(spec.command, spec.args ?? [])}`;
@@ -40,11 +49,35 @@ function rejectIfShellFailed(
   return undefined;
 }
 
-export async function runShellAction(spec: ShellActionExecution): Promise<ShellActionResult> {
-  const cwd = spec.cwd ?? process.cwd();
-  const args = spec.args ?? [];
-  const startMs = Date.now();
-  const child = spawn(spec.command, args, {
+function settleShellResult(
+  spec: ShellActionExecution,
+  args: string[],
+  result: ShellActionResult,
+  timedOut: boolean,
+  aborted: boolean,
+): Error | undefined {
+  if (aborted && !timedOut) {
+    return new InterruptedError();
+  }
+  return rejectIfShellFailed(spec, args, result, timedOut);
+}
+
+function scheduleShellTimeout(
+  timeoutMs: number | undefined,
+  onTimeout: () => void,
+): NodeJS.Timeout | undefined {
+  if (timeoutMs == null || timeoutMs <= 0) {
+    return undefined;
+  }
+  return setTimeout(onTimeout, timeoutMs);
+}
+
+function spawnShellChild(
+  spec: ShellActionExecution,
+  cwd: string,
+  args: string[],
+): ChildProcessWithoutNullStreams {
+  return spawn(spec.command, args, {
     cwd,
     env: {
       ...process.env,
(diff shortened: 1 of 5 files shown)


steipete:
Maintainer triage: CLOSE recommended for the current implementation. Independent built-CLI proof confirms that interruption still leaks both resistant direct children and shell descendants.

The proof launched a real flow through this PR's `dist/cli.js`, waited for a Node child to write its PID, sent SIGINT to the CLI, and checked the recorded processes after 1600 ms:

```text
resistant-direct-child:
  CLI exit=130
  childAliveAfterGrace=true
shell-wrapper-descendant:
  CLI exit=130
  wrapperAliveAfterGrace=false
  childAliveAfterGrace=true
```

Environment: macOS, Node 26.8.1, isolated synthetic HOME/cwd/TMPDIR, no provider or network. The fixture's final cleanup killed and verified absence of every known test process.

The root cause is the remaining ownership gap: the interrupt path aborts but does not await command teardown before `src/cli-core.ts` calls `process.exit(130)`. The unref'ed escalation timer cannot then run. Separately, `child.kill()` targets the wrapper PID, so descendants survive even when the wrapper exits.

A complete replacement needs one runtime-owned, awaited command-tree cleanup boundary shared by timeout and interruption. It must wait only for owned process teardown, not for arbitrary user `exec`/`parse` callbacks that might never settle. The current green tests do not establish those guarantees. No contributor changes are requested, and the PR is left open for the orchestrator's disposition.

steipete:
Thanks @SebTardif for working on this. The interruption/timeout cleanup problem is real, but I’m closing this implementation because the [built-CLI proof](https://github.com/openclaw/acpx/pull/530#issuecomment-5476312676) still leaves both a SIGTERM-resistant direct child and a shell-wrapper descendant alive 1.6 seconds after the CLI exits with code 130.

Cancellation needs a shared, awaited process-tree cleanup boundary: finish teardown and escalation for owned processes before CLI exit, without waiting indefinitely for user callbacks. Killing only the wrapper PID and scheduling an unref’ed escalation timer does not provide that guarantee. A corrected design with interruption and timeout proof for resistant children and descendants is very welcome.

PR #531, opened 2026-08-29
feat(flows): add opt-in shell capture limits

SebTardif:
Flow shell actions accumulate captured stdout and stderr without a bound. Add optional `maxBufferBytes` per action so callers can bound each stream independently. Omission retains unlimited capture; zero permits empty output only. Values count UTF-8 bytes and must be non-negative safe integers.

Overflow discards further output and uses the current process-tree cancellation owner. The caller receives the buffer error only after cleanup finishes. Existing stdin handling, successful wrapper-exit behavior, deadlines and cancellation precedence are preserved. This replaces the branch's older separate kill implementation and avoids truncating a multibyte character into a larger replacement character.

Thanks @SebTardif for identifying and demonstrating the issue. Release notes are consolidated in the final notes PR.

Validation: 27 focused shell tests, docs checks, clean local autoreview, and a separate host importing built `dist/flows.js` through `defineFlow`, `shell` and `FlowRunner`. The live run captured 2 MiB with no cap, rejected both capped streams, kept the host alive, and verified a SIGTERM-resistant descendant was gone before rejection. `pnpm run check` passed with 1,006 passing tests (two skips) and 147 passing coverage tests. Branch autoreview is clean through P2. Prepared head: `a6ce049bb29036c59ff730822e0c77b7b6efe56c`. CI: https://github.com/openclaw/acpx/actions/runs/34102635671.


steipete:
Verified `a6ce049bb29036c59ff730822e0c77b7b6efe56c`: `pnpm run check` passed (1,006 tests, two skips; 147 coverage tests), docs checks passed, and isolated branch autoreview is clean through P2.

A separate host used built `dist/flows.js` through `defineFlow`, `shell` and `FlowRunner`: uncapped capture retained exactly 2,097,152 bytes; explicit caps rejected stdout and stderr; a SIGTERM-resistant descendant was confirmed gone before the rejection reached the host. The host stayed alive. Focused regression coverage also checks split UTF-8, zero/exact limits, invalid limits before spawn and cancellation precedence.

CI for this head: https://github.com/openclaw/acpx/actions/runs/34102635671. Release notes are consolidated in the final notes PR.

PR #543, opened 2026-09-03
feat(runtime): allow external ACP process launchers

devSejung:
## Summary
- add a public optional ACP agent process launcher to the embedded runtime
- preserve logical agent command metadata while allowing an SSH or container stdio transport
- expose filesystem and terminal capability switches through runtime options
- retain the existing local spawn path when the launcher declines

## Verification
- pnpm run typecheck
- pnpm run build:test
- node --test --test-name-pattern=processLauncher dist-test/test/client.test.js


steipete:
Thanks @devSejung for identifying this embedding gap. We are keeping the external-launcher API deferred and closing this proposal as not planned, consistent with the decision recorded in https://github.com/openclaw/acpx/pull/619.

The disabled-capability enforcement issue is already fixed. The remaining concern is the public transport contract: this callback returns a Node child process while acpx owns local PID admission and descendant cleanup, and the launch command and session cwd are resolved for the controller. The callback does not establish remote teardown or workspace mapping, and it predates the current startup and cleanup guarantees.

Custom agent argv can already launch an SSH/container stdio wrapper through the existing local-child lifecycle. That wrapper becomes the command/session identity; it is not a transparent replacement for a built-in harness. Exposing runtime `fs`/`terminal` options is a useful separate improvement, but does not require adopting a launcher API.

PR #552, opened 2026-09-07
feat(runtime): support transient agent environment overlays

### LINKED ISSUE #517: Transient per-runtime agent child environment (non-persisted `agentProcessEnv`)
## What I'm asking for

An optional, **non-persisted** environment input on `AcpRuntimeOptions` that is merged into every agent child the runtime spawns:

```ts
type AcpRuntimeOptions = {
  // …
  /**
   * Merged into every agent child this runtime spawns, after session env,
   * never overriding protected auth keys. Never persisted.
   */
  agentProcessEnv?: Record<string, string>;
};
```

## Why the existing inputs don't cover it

I need to tell an ACP adapter **which build of its underlying CLI to run**, per runtime, without that choice outliving the runtime. Concretely: the Claude adapter picks its binary from `CLAUDE_CODE_EXECUTABLE` in its own process environment.

The two available routes both have side effects I can't accept:

1. **`SessionAgentOptions.env`** is persisted into the session record and reloaded. In 0.13.1, `persistSessionOptions()` writes it and `sessionOptionsFromRecord()` reads it back (`live-checkpoint-BSIrfgVo.js:5264`, `:5276`, `:5284`), and it appears in the record shape as `SessionAcpxState.session_options.env` (`session-options-DwRDODlr.d.ts:232`). An absolute executable path is host layout — it stops being true when a build moves — so persisting it means a stored record can later point a child at a binary that is gone or different.

2. **Mutating `process.env`** binds every other child the process spawns, not just this runtime's.

What's missing is a per-runtime, transient input: applied to the children this runtime spawns, and forgotten with it.

## Use case

I'm binding one exact provider executable to a durable session. A native `claude` UI creates a conversation, and later an ACP client has to attach to *that* conversation. If the ACP child runs a different Claude build than the one that created the session, it accepts the session id and silently disagrees about what it names — we hit exactly that: one build created a session, another was asked to resume it, and produced an empty conversation. (Related, from the adapter side: agentclientprotocol/claude-agent-acp#1019.)

So we observe the executable — canonical path, SHA-256, reported version — retain the version and digest, and need to hand the *path* to the matching ACP child for that one attachment only. The path must reach the child and nothing else.

## Suggested shape

*Edited 2026-08-23: the merge originally proposed here was a raw spread after `buildAgentEnvironment`, which would have bypassed the protected-auth-key guard — the review comment below is right about that. Amended so the merge goes through the same guard as session env.*

Threading it takes four small changes:

- `AcpRuntimeManager.createClient()` merges `this.options.agentProcessEnv` into every client it creates — doctor, ensure, reconnect, control and turn all funnel through here.
- `createProbeClient()` builds its own options and never reaches the manager, so it needs it threaded explicitly.
- `buildAgentSpawnOptions(cwd, authCredentials, sessionEnv)` takes a fourth argument and passes it through to `buildAgentEnvironment` (`src/acp/auth-env.ts`), where it merges **after** `sessionEnv` via the existing machinery: keys in `protectedAuthEnvKeys` are skipped, and assignment goes through `assignSessionEnv` so the win32 case-collision handling applies.

```js
// buildAgentEnvironment(authCredentials, sessionEnv, agentProcessEnv)
if (agentProcessEnv) {
  for (const [key, value] of Object.entries(agentProcessEnv)) {
    if (typeof value !== "string" || protectedAuthEnvKeys.has(protectedEnvKey(key))) {
      continue;
    }
    assignSessionEnv(env, key, value);
  }
}
```

The precedence contract is then: **protected auth keys > `agentProcessEnv` > `sessionOptions.env` > inherited `process.env`**. That retains the property we need — a caller that has bound an executable wins over whatever the adapter would otherwise resolve for itself — since the keys we pass (`CLAUDE_CODE_EXECUTABLE`) are never auth keys. A transient env silently swapping credentials is nothing we want either.

The property we depend on is that it is **never** copied into `SessionAgentOptions`, a session record, an event, a status, a diagnostic or a serialized store, and never applied to the parent process.

## Current workaround

We vendor a source snapshot of `acpx@0.12.0` carrying this patch (in its earlier raw-spread form — harmless there only because we pass a single non-auth key), plus a verifier that proves it is the only behavioural difference from the pristine upstream bytes. That's a snapshot we'd very much like to delete — our removal ticket is gated on a released `acpx` supplying this seam.

Happy to open a PR implementing the above against `main` if you'd like it in this shape, or to adjust the naming/placement to whatever fits your API better. Also happy to move this to a Discussion — `CONTRIBUTING.md` points feature requests there, but Discussions look disabled on this repo, and #387 and #502 suggested issues are an accepted venue.

steipete:
Embedded hosts need child-process settings that last for one runtime without changing `process.env` or saving machine-specific paths into session records. Add optional `agentProcessEnv`, snapshotted at runtime construction and forwarded through the shared client builder and independent probe path.

The overlay uses existing credential protection and Windows environment-name handling. Protected authentication values win, then runtime values, saved session values, and inherited environment. Invalid entries fail before spawn without echoing values. Adapter output remains adapter-owned and may itself contain anything the adapter chooses to emit.

Fixes #517. Thanks @taras and @coding-ax for the concrete embedding use cases. Changelog credit is in the consolidated notes PR.

Validation: format, typecheck, lint, package/viewer builds and docs checks passed; the full Node 26 suite passed 1,005 tests with two skips, and all 148 coverage tests passed. Isolated branch autoreview is clean through P2. CI passed on `8045e6474f9ffe99e5fefd70e801280f4e9b77a5`: https://github.com/openclaw/acpx/actions/runs/34100628378. Local Node 24 viewer tests intermittently timed out or retained a worker even after a fresh-temp retry; Node 26 and CI both completed. A separate host imported built `dist/runtime.js`. The live host observes three real child starts: probe and first session receive the immutable runtime overlay; a new runtime reconnects using the original saved session value. Store inspection and parent-environment assertions confirm no overlay persistence or parent mutation.

For this batch, maintainer direction is to keep all changelog edits in https://github.com/openclaw/acpx/pull/554 and land that PR last. This overrides the normal per-PR changelog placement rule; the consolidated section includes the feature and contributor credit.

Fresh copied output from a build of the exact head (`8045e6474f9ffe99e5fefd70e801280f4e9b77a5`) extracted with `git archive`. The child environment values below are synthetic; the runtime host asserted the observed values, recursively checked the session store, and checked the parent environment.

```json
{"probe":"runtime overlay","firstSession":"construction snapshot","reconnect":"saved session value","persistedOverlay":false,"parentChanged":false,"childStarts":3,"childEnvironments":[{"overlay":"runtime-only-proof-value"},{"overlay":"runtime-only-proof-value"},{"overlay":"saved-session-value"}]}
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -213,3 +213,6 @@ Then `acpx ci-bot 'run sanity checks'` resolves through the registry without any
 - [Custom agents](custom-agents.md) — `--agent` escape hatch and unknown positional names.
 - [Permissions](permissions.md) — `defaultPermissions` and non-interactive policy.
 - [Output formats](output-formats.md) — `format` default and `--json-strict`.
+
+Embedded hosts can also supply a [transient runtime environment](runtime-environment.md)
+without persisting child-process settings in a session.
--- a/docs/runtime-environment.md
+++ b/docs/runtime-environment.md
@@ -1,0 +1,32 @@
+# Transient runtime environment
+
+Embedded clients can set `AcpRuntimeOptions.agentProcessEnv` to a map of strings
+for the agent children owned by that runtime. ACPX snapshots the map when
+`createAcpRuntime()` is called. Later mutations of the caller's map do not affect
+that runtime. The same overlay applies to probes, new sessions, controls and
+reconnections, including sessions loaded from an existing store.
+
+Precedence is protected authentication values, then `agentProcessEnv`, then
+persisted `sessionOptions.env`, then the inherited process environment. Windows
+uses the same case-insensitive collision handling as session environment values.
+The overlay does not change the parent process environment.
+
+This is trusted embedding-host configuration: executable, loader and other
+process settings can change what the child runs. ACPX does not sanitize arbitrary
+environment variables. Existing credential protection remains in effect.
+Invalid environment entries are rejected without echoing their values.
+
+ACPX does not copy the overlay into session records, events, status or diagnostics.
+An adapter can still echo its own environment into output; this option cannot
+prevent that. A new runtime must supply its own overlay, and a previously saved
+session environment retains its existing persistence behavior.
+
+```ts
+const runtime = createAcpRuntime({
+  cwd,
+  sessionStore,
+  agentRegistry,
+  permissionMode: "approve-reads",
+  agentProcessEnv: { ADAPTER_EXECUTABLE: "/opt/agent/bin/agent" },
+});
+```
--- a/src/acp/auth-env.ts
+++ b/src/acp/auth-env.ts
@@ -79,10 +79,27 @@ function promotePrefixedAuthEnvironment(env: NodeJS.ProcessEnv): Set<string> {
   return protectedKeys;
 }
 
+function validateAgentProcessEnv(agentProcessEnv: Record<string, string> | undefined): void {
+  for (const [key, value] of Object.entries(agentProcessEnv ?? {})) {
+    if (
+      typeof value !== "string" ||
+      key.includes("=") ||
+      key.includes("\u0000") ||
+      value.includes("\u0000")
+    ) {
+      throw new Error(
+        "Invalid agentProcessEnv: environment entries cannot contain NUL or names with '='",
+      );
+    }
+  }
+}
+
 function buildAgentEnvironment(
   authCredentials: Record<string, string> | undefined,
   sessionEnv: Record<string, string> | undefined,
+  agentProcessEnv: Record<string, string> | undefined,
 ): NodeJS.ProcessEnv {
+  validateAgentProcessEnv(agentProcessEnv);
   const env: NodeJS.ProcessEnv = { ...process.env };
   const protectedAuthEnvKeys = promotePrefixedAuthEnvironment(env);
   if (authCredentials) {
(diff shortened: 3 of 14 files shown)


steipete:
Verified `8045e6474f9ffe99e5fefd70e801280f4e9b77a5` with 1,005 passing full-suite tests on Node 26 (two skips), 148 passing coverage tests, 44 focused environment/probe checks, formatting, typecheck, lint, builds and docs checks. Isolated branch autoreview is clean through P2. CI is green: https://github.com/openclaw/acpx/actions/runs/34100628378.

A separate host imported built `dist/runtime.js` and observed three real ACP child starts. The probe and first session received the construction-time overlay; a new runtime without an overlay reconnected with the original saved session value. Recursive store inspection found no overlay value and the parent environment stayed unchanged. Invalid environment values fail without being echoed into errors. All proof values were synthetic.

Local Node 24 replay-viewer tests intermittently timed out or retained their worker, including after the required fresh-private-temp retry. The full supported Node 26 run and hosted CI completed successfully.

steipete:
Rebuilt exact head `8045e6474f9ffe99e5fefd70e801280f4e9b77a5` from `git archive` and reran the external host against built `dist/runtime.js`. Copied program output:

```json
{"probe":"runtime overlay","firstSession":"construction snapshot","reconnect":"saved session value","persistedOverlay":false,"parentChanged":false,"childStarts":3,"childEnvironments":[{"overlay":"runtime-only-proof-value"},{"overlay":"runtime-only-proof-value"},{"overlay":"saved-session-value"}]}
```

The three values are the actual child-start audit rows: probe, first session, then a new runtime reconnecting without the transient overlay. The host asserts those rows, absence of the transient value from the complete store, and absence of parent mutation. All values are synthetic.

For this batch, maintainer direction is to keep all changelog edits in https://github.com/openclaw/acpx/pull/554 and land that PR last. This overrides the normal per-PR changelog placement rule; the consolidated section includes the feature and contributor credit.

taras:
Awesome! 

Thank you for getting this sorted as releasing it in 0.15. 

I can now remove my vendored version of ACPX from XMD.

PR #514, opened 2026-08-22
feat(runtime): expose process lifecycle hooks to embedding hosts

MertBasar0:
Related: #499

Context: openclaw/openclaw#128006

## What Problem This Solves

Embedding hosts can launch ACP agents through `AcpClient` or `AcpRuntime`, but they currently have no supported way to durably record process ownership around the exact spawn boundary. Hosts that need crash recovery or process-lease reconciliation must either wrap ACPX internals or accept a gap between preparing a launch and learning the child PID.

## Why This Change Was Made

This adds an optional lifecycle contract with immutable, correlated events for pre-spawn admission, successful spawn, spawn failure, and process exit. Runtime-created clients also receive stable session/probe scopes.

The two admission hooks are deliberately awaited: rejecting before spawn prevents launch, while rejecting after spawn terminates the child before returning the error. Failure and exit observers remain best-effort and cannot mask the underlying process outcome. Exit notification is ordered after spawned admission settles, avoiding a stale-lease race when a child exits quickly.

This is an embedding seam only. It does not add persistence, process-tree cleanup, or host-specific lease policy to ACPX, and therefore does not claim to close #499.

## User Impact

Embedding developers can now implement durable host-owned launch tracking without wrapping ACPX spawn internals. Existing clients are unaffected because every lifecycle hook is optional.

## Evidence

- Full `pnpm run check` and `pnpm run check:docs` pass in a clean Linux checkout: formatting, TypeScript checks, type-aware lint, package and replay-viewer builds, 997 repository tests (996 passed, 1 skipped), 145 coverage tests, and documentation formatting/lint/site build.
- Ten focused lifecycle/runtime regression tests pass, covering event ordering, admission rejection, child cleanup, spawn-failure correlation, a never-settling failure observer, pre-observer exit replay, early exit during successful spawned admission, initialization exit, runtime-session scope, and probe scope.
- The P2 follow-up makes `onSpawnFailed` nonblocking while preserving synchronous invocation, correlation data, and logging for synchronous throws and asynchronous rejections.
- The P1 follow-up installs startup-exit observation before awaited spawned admission and checks already-recorded child exit state, preventing startup from hanging when the agent exits while a successful admission hook is still pending.
- The latest P1 follow-up replays an exit already recorded before lifecycle observer attachment through the same admission-ordering barrier, preventing stale host process records; its regression test fails on the preceding commit and passes on the current head.
[...]

PR #553, opened 2026-09-07
feat(runtime): expose optional process lifecycle admission hooks

### LINKED ISSUE #499: Follow-up on #433: ACP bridge lifeline / process cleanup
Following up on #433 after your closing comment.

The pipe-lifeline approach addressed a real problem I’m still running into: I invoke ACPX sessions heavily and sometimes in sequence, and leaked bridge/model/MCP processes eventually accumulate and start causing reliability issues with subsequent sessions.

#433 already went quite far on the implementation and regression coverage, but I understand the concern around the predictable native-helper cache and why you didn’t want that execution boundary in `~/.acpx/native`.

Before taking another pass at it, I’d rather align with you on the architecture you’d be comfortable having in ACPX.

Do you have a preferred direction for keeping the helper inside the package boundary and handling the integrity/trust side of it? Also useful to know if there are approaches you definitely *don’t* want here.

I’m happy to rework the implementation from current `main` once the intended shape is clearer.

For context, having this upstream matters quite a bit for my use case: carrying the lifecycle fix separately while ACPX moves quickly means repeatedly rebasing it, whereas this is behavior I rely on when running a lot of ACPX sessions.

Related: #433 (and the earlier #432).

steipete:
Embedding hosts need to record launch ownership at the actual ACP process boundary. Preserve the optional, correlated lifecycle hooks from https://github.com/openclaw/acpx/pull/514 on current main, with explicit documentation of admission and observer semantics.

`onBeforeSpawn` and `onSpawned` are awaited. Rejection prevents launch or cleans up the spawned child before returning. Hosts own admission timeouts and recovery; `close()` does not cancel a pending admission hook. Failure and exit observers are best-effort and cannot replace the underlying outcome. Exit notification waits for spawned admission to settle and replays an exit recorded before observer attachment.

This carries @MertBasar0's implementation and regression coverage forward because the contributor branch does not allow maintainer edits. It adds no persisted process leases or abrupt-death/process-tree cleanup guarantee and does not close #499. Contributor credit is retained in the commit and consolidated release notes.

Verification includes the full repository checks, docs checks, eight focused lifecycle cases, isolated review, and an external host importing built `dist/runtime.js`. Real child processes demonstrate ordered immutable events, pre/post-spawn admission rejection and cleanup, exit 17 during slow successful admission, a non-settling failure observer that does not block, and runtime-probe scope. `pnpm run check` passed with 1,030 passing tests (two skips) and 148 passing coverage tests. Branch autoreview is clean through P2. Exact head: `940f1a69c79ba4e8f826977aa7c5c5d1522b7e9c`. CI: https://github.com/openclaw/acpx/actions/runs/34110157976.

For this batch, maintainer direction is to keep all changelog edits in https://github.com/openclaw/acpx/pull/554 and land that PR last. This overrides the normal per-PR changelog placement rule; the consolidated section includes the feature and contributor credit.

Rebased onto main at `836580394b6e675321b6924b1f3dc79287fb1aca`. Conflict resolution retains both environment-overlay and lifecycle assertions. Additional built-runtime proof verified nine correlated lifecycle events across a probe, initial session and reconnect while preserving transient environment snapshots; built runtime/compare metadata and ACP overflow error checks also pass.


steipete:
Verified `9d09054e966d69af3fd23b6560315eea256245c0`: full `pnpm run check` passed (1,009 tests, two skips; 147 coverage tests), docs checks and eight focused lifecycle cases passed, and isolated branch autoreview is clean through P2.

An external host imported built `dist/runtime.js` and drove real OS processes. It verified immutable correlated `before → spawned → exit` events; pre-spawn admission rejection; post-spawn rejection with the PID gone before return; exit 17 during slow successful admission with ordered exit delivery; a non-settling failure observer that did not delay spawn failure; and runtime-probe scope. All processes and fixture values were synthetic.

CI: https://github.com/openclaw/acpx/actions/runs/34105237587. This preserves @MertBasar0's contribution from #514 and does not resolve the separate cleanup policy in #499.

steipete:
Rebased onto main at `836580394b6e675321b6924b1f3dc79287fb1aca`. New head: `940f1a69c79ba4e8f826977aa7c5c5d1522b7e9c`. The three test conflicts retain both environment and lifecycle assertions. Human contributor credit is preserved; changelog edits remain in #554.

`pnpm run check` passed: 1,030 tests, two skips, and 148 coverage tests. `pnpm run check:docs` passed. Isolated branch autoreview is scoped-clean through P2.

An external host imported built `dist/runtime.js` and drove real child processes through ordered immutable lifecycle events, both admission rejection paths, early exit 17, nonblocking failure observation and runtime-probe scope. A combined environment/lifecycle run verified nine correlated events across probe, session and reconnect launches, snapshot isolation, and no overlay persistence or parent-environment mutation. Built runtime/compare preserved opaque metadata; built CLI checks accepted a 9 MiB uncapped frame and reported capped overflow during startup and prompts.

Synthetic proof output:

```json
{"normalOrder":["before","spawned","exit"],"admissionRejections":"cleaned","earlyExit":17,"exitOrder":["start","end","exit"],"nonsettlingFailureObserver":"nonblocking","probeScope":"runtime-probe"}
{"correlatedLifecycleEvents":9,"probe":"runtime overlay","firstSession":"construction snapshot","reconnect":"saved session value","persistedOverlay":false,"parentChanged":false,"childStarts":3}
```

Exact-head CI: https://github.com/openclaw/acpx/actions/runs/34110157976. This does not close #499.

MertBasar0:
Thanks for carrying this forward and preserving the contributor credit. I had overlooked that 'Allow edits by maintainers' was disabled on the original fork PR, so I appreciate you handling the extra integration work. I will make sure it is enabled on dedicated contribution branches going forward.

PR #558, opened 2026-09-08
fix(runtime): let hosts withhold secret env keys from session records

Rick7C2:
fix(runtime): let hosts withhold secret env keys from session records

## The problem

`SessionAgentOptions.env` is documented as persisted to the session record, with the
guidance to route credentials through `authCredentials` instead
(`src/runtime/engine/session-options.ts`):

> Do not put secrets here; use `authCredentials` for credentials.

That guidance is unreachable for runtime embedders. `AcpRuntimeOptions`
(`src/runtime/public/contract.ts`) exposes `sessionOptions.env` but has no
`authCredentials` equivalent — it is reachable only from the CLI and the `flows` API
(`FlowRunnerOptions`, `src/flows/types.ts:349`).

A host embedding `createAcpRuntime` with per-agent credentials therefore has no
compliant channel short of adopting the heavier flows abstraction, and its credentials
land in `acpx.session_options.env` in cleartext on disk.

## The fix

An opt-in `secretEnvKeys` on `AcpRuntimeOptions` naming env keys to withhold from the
persisted record. The keys are still passed to the spawned agent; only the on-disk copy
is dropped. **Behaviour is byte-identical when the option is omitted** — `undefined`
short-circuits every new code path, so no existing consumer is affected.

### Why omit rather than placeholder

On resume the record is the *sole* source of session env — `manager.ts` and
`connected-session.ts` both build client options from `sessionOptionsFromRecord` with no
live fallback — and `buildAgentEnvironment` layers session env *over* `process.env`. A
placeholder would be assigned to the child as a literal value and break the very launch
it was meant to protect. Omission lets the host's freshly-supplied value stand.

## A second bug, fixed here too

The one-shot reuse gate compares stored options against live ones with
`isDeepStrictEqual`. Once any key is withheld that comparison can never match, silently
disabling one-shot session reuse with no error. This commit compares against
`persistableSessionOptions` — the view that survives the round trip — including the case
where *every* env key is withheld, where the record carries no `session_options` block at
all and the round-trip view is `undefined` rather than `{}`.

## Documented consequence

Withheld keys do not participate in the reuse comparison, so changing only a withheld
value no longer forces a fresh session. Rotate by using a new session key if a rotated
credential must reach a new child. This is noted on the option's doc comment.

## Verification

Against a clean clone at `9ace847`:

- `pnpm run typecheck` / `lint` / `format:check` / `build` / `build:test` — all clean
- `node --test dist-test/test/runtime-manager.test.js` — **75/75** (6 new tests)

Mutation-checked rather than merely green:

| Mutation | Result |
|---|---|
| Revert only the empty-object one-liner | 74 pass / 1 fail — the new all-secret reuse test |
| Neuter the filter itself (`false && isSecretEnvKey`) | 73 pass / 2 fail — both persistence tests |

Neither test passes vacuously.

Pre-existing failures elsewhere in the suite reproduce identically on an unmodified
`9ace847` tree and are unrelated to this change.


steipete:
Closing as superseded by the runtime-only `AcpRuntimeOptions.agentProcessEnv` overlay already on main in https://github.com/openclaw/acpx/commit/ab5c2b3c0cd3434db8b2781a4ad76688ed62b227. The supported contract is documented in [docs/runtime-environment.md](https://github.com/openclaw/acpx/blob/main/docs/runtime-environment.md).

I verified the built `dist/runtime.js` against a separate Node ACP agent over stdio JSON-RPC. The first runtime and a replacement runtime each supplied a different synthetic overlay; the child returned the expected SHA-256 digest at final prompt I/O. The replacement used `session/load`. Both child processes received the intended overlay, the parent environment stayed unchanged, mutating the caller's map after runtime construction had no effect, and the saved session record contained neither transient value. The deliberately persisted `sessionOptions.env` value remained intact.

This avoids adding a second secret-key classification API and per-session transient cache. Hosts should supply credentials through `agentProcessEnv`, recreate the runtime when rotating its snapshot, and use separate runtimes where credentials differ. This does not retroactively scrub previously persisted session environments or promise that an adapter cannot echo its environment. Thanks for identifying and testing this use case.

PR #570, opened 2026-09-12
fix(replay): preserve message identities across transcript projection

steipete:
## What Problem This Solves

The replay viewer emits message-ID changes while a session bundle is idle because each transcript projection creates new random user IDs.

## User Impact

Repeated reads of unchanged history produce no patches. New text still streams normally, and existing checkpoint IDs remain authoritative.

## Why This Change Was Made

Separate identity allocation from conversation projection: live writers keep UUID defaults, while replay assigns IDs from the session bundle and event sequence for post-checkpoint messages. Both prompt submissions and inbound user-message chunks use the same policy. No wire or stored-bundle schema changes.

## Evidence

- Two regression cases failed before the fix, differing only in random user IDs. They now pass and cover repeated reads, transcript growth, and persisted checkpoint IDs.
- 15 focused replay/conversation tests passed.
- A standalone built-server proof used actual HTTP/WebSocket transport and filesystem bundles: `{"polls":11,"idlePatches":0,"appendedMarker":true,"stableUserIds":true}`. The server remained quiet during unchanged reads, then emitted the expected text append.
- Isolated P0–P2 Codex autoreview: scoped-clean.

- `pnpm run check` passed: 1,046 tests passed, two existing platform skips; 148 coverage tests passed. Documentation checks also passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Repo: https://github.com/openclaw/acpx
 
 ### Fixes
 
+- Replay viewer: preserve user-message identities across repeated transcript projection so idle polling does not emit spurious patches.
 - Tooling: prevent malformed TOML configuration from hanging documentation lint by overriding the vulnerable `smol-toml` pin with 1.7.2 (GHSA-7w5x-hrqm-74c2).
 
 ## 0.15.1 - 2026-09-07
--- a/docs/2026-03-31-flow-replay-live-transport.md
+++ b/docs/2026-03-31-flow-replay-live-transport.md
@@ -64,6 +64,9 @@ re-decided during coding:
 - The server computes patches from **semantic viewer state**, not by patching
   storage files directly.
 - The on-disk bundle format stays unchanged in this work.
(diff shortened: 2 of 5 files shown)


steipete:
Landed as https://github.com/openclaw/acpx/commit/42459e49de3d6959f6d5d28250cf1d327d76fcf6.

[Exact-head CI](https://github.com/openclaw/acpx/actions/runs/34721690871) passed on Node 22/24/26, including coverage, mutation, docs, conformance, and CodeQL. Local validation passed 1,046 tests with two existing platform skips plus 148 coverage tests. Both new deterministic regressions failed before the fix and passed afterward. The built production live-sync server observed 11 filesystem reads over actual HTTP/WebSocket transport without idle patches, then delivered the expected text append with stable identities.

Projection now accepts caller-owned identities; live writers retain UUID defaults, and checkpoint IDs remain authoritative. Isolated P0–P2 review found no actionable defect. The changelog stays ordered most-interesting-first as explicitly requested for this sweep, overriding the generic append-only preference.

PR #561, opened 2026-09-10
feat: add fx as a built-in agent

wtfsayo:
## What Problem This Solves

fx currently needs the raw `--agent 'fx acp'` form. This adds the built-in `acpx fx …` shortcut.

## User Impact

The shortcut runs the installed fx ACP server with its existing provider configuration. Model selection, session reuse, raw overrides, and configured commands use the existing acpx paths. Choose the provider with fx's own CLI before launch.

## Why This Change Was Made

One entry in the existing agent-definition registry supplies the command, argv, discovery, and ordering. Three identical native-agent integration tests and launch helpers are consolidated into one strict fixture, retaining MCode, Pool, and ZeroClaw coverage while adding fx. The result is 89 fewer lines overall, with one new production line.

## Evidence

- 27 registry tests and four native-launch integration cases pass. The shared fixture still rejects a missing or wrong ACP argument on both platform branches.
- Independent source-blind validation passed 28 assertions: exact argv, help, overrides, prompt integrity, standard acpx metadata, model-selector precedence, error handling, and persistence across a server-process restart.
- Official fx 0.0.10 was checked against its release SHA256 and launched through the built shortcut in isolated state. ACP initialize returned the expected provider-credential requirement with actionable `fx login`/`fx setup` guidance. No sign-in or credentials were used, and fresh authenticated completion is not claimed.
- The contributor's existing fx 0.0.8 Grok evidence covers completed turns, model selection, and session persistence. That proof remains separate from the fresh native startup/auth-boundary probe.
- Independent review found no actionable issues on the final candidate; the provider setup command is verified against the native CLI. Full Linux code/docs checks passed: 1,195 tests passed with two platform skips, plus all 171 coverage-target tests. The agent guide and published skill passed their additional formatting/Markdown gates. Public exact-head GitHub CI is checked before landing.

Thanks @wtfsayo for the implementation and live evidence. This preserves the original PR and contributor credit while adapting it to the current registry owner.

PR #571, opened 2026-09-12
fix(acp): select model controls ahead of provider selectors

steipete:
Related: https://github.com/openclaw/acpx/pull/561

## What Problem This Solves

`--model` can reject a valid model—or target a provider selector—when an adapter advertises provider and model controls in the same model category.

## User Impact

acpx selects the categorized `model` control when present, preserves custom categorized controls, and retains legacy ID-based fallback. Provider configuration stays separate from model selection.

## Why This Change Was Made

The parser previously returned the first valid candidate. Rank validated candidates using both existing signals: categorized `model`, other model-category controls, then legacy `id=model`. Stable ties and grouped-option parsing remain intact. This isolates the generic parser defect reported by @wtfsayo from the separate profile and startup-model proposals.

## Evidence

- Three parser regressions failed before the fix; all ten model-support tests now pass. Cases cover both control orders, custom/legacy precedence, malformed candidates, and grouped options.
- The real-process CLI regression failed before and passed after.
- Built CLI before: `--model smart-model` failed with `Available models: fixture-provider`. The same command afterward sent exactly one `session/set_config_option` with `configId=model`, `value=smart-model`, and completed with `end_turn`.
- Isolated P0–P2 Codex autoreview: scoped-clean.
- Release notes follow the maintainer-requested most-interesting-first order for this maintenance sweep.
- `pnpm run check`: 1,050 passed, two existing skips; coverage suite 148/148 passed. `pnpm run check:docs`: format, lint, and site build passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Repo: https://github.com/openclaw/acpx
 
 ### Fixes
 
+- Model selection: prefer the actual model control when an adapter also groups provider controls under the model category, while retaining custom model controls and legacy fallback. Thanks @wtfsayo.
 - Replay viewer: preserve user-message identities across repeated transcript projection so idle polling does not emit spurious patches.
 - Tooling: prevent malformed TOML configuration from hanging documentation lint by overriding the vulnerable `smol-toml` pin with 1.7.2 (GHSA-7w5x-hrqm-74c2).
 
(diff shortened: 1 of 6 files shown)


steipete:
Landed the generic model/provider selector fix with @wtfsayo credited. Model-category controls now use stable precedence without dropping custom IDs or the legacy fallback; the broader startup/profile proposals remain open.

Proof: three parser regressions and the real-process CLI case failed before the change; the built CLI now sends exactly one `session/set_config_option` to `configId=model` and completes the turn. Full checks passed (1,050 tests, two existing Windows-only skips, coverage 148/148), docs checks passed, and isolated P0–P2 review was scoped-clean. Exact-head CI: https://github.com/openclaw/acpx/actions/runs/34722430977.

PR #576, opened 2026-09-15
fix(sessions): export large histories without stack overflow

steipete:
Exporting a large session event segment crashed with `Maximum call stack size exceeded`: the reader spread every parsed event into one `Array.push` call. Export now appends parsed messages directly, preserving the complete history in rotated-to-active segment order.

The regression uses 300,000 distinct events across two segments and verifies the complete ordered archive. A separate built-CLI reproduction with 150,000 synthetic events exited 1 before the fix and exits 0 afterward with every event in its original position.

The full suite also exposed an existing shutdown-test race. PID creation happens before ACP initialization, so signalling at that point did not establish the active prompt the assertion expected. The test now waits for a streamed `prompt-ready` marker before SIGTERM. A generated fixture with four seconds of delayed initialization verified cancelled completion, bridge exit, and lease retention/release. Production shutdown behavior is unchanged.

Validation:

- Session export/import regression suite: 19 passed.
- Built CLI: all 150,000 fixture events exported in exact order.
- Delayed-startup shutdown fixture passed.
- `pnpm run check`: 1,069 passed, two skipped; all 152 coverage tests passed. The configured coverage target cleared all 85% gates (95.06% lines/statements, 88.71% branches, 96.67% functions).
- `pnpm run check:docs`: passed.
- Independent isolated review: no actionable P0–P2 findings.

Includes the session-guide clarification and an Unreleased changelog entry.

<details>
<summary>Built-CLI proof: commands, synthetic fixture, and captured results</summary>

Node.js `v24.21.0` on macOS. The same synthetic fixture was used on base `bfd7c92fa180016f98776d06a64cef37306f642a` and PR head `6fea12b940d1201d97635d49c71a4f72fc6f60c4`, after `pnpm run build`. Only local temporary paths are normalized below as `$FIXTURE` / `<fixture>`.

The fixture is an inactive session with one 150,000-message NDJSON segment. It does not launch an adapter or use an authenticated account. Recreate it with:

```sh
export FIXTURE="$(mktemp -d)"
python3 - <<'PY'
import json, os, pathlib
root = pathlib.Path(os.environ['FIXTURE'])
cwd = root / 'workspace'
sessions = root / '.acpx' / 'sessions'
sessions.mkdir(parents=True)
cwd.mkdir()
timestamp = '2026-09-14T00:00:00.000Z'
record = {
    'schema': 'acpx.session.v1',
    'acpx_record_id': 'large-history',
    'acp_session_id': 'synthetic-backend',
    'agent_command': 'synthetic-agent',
    'cwd': str(cwd), 'name': 'large',
    'created_at': timestamp, 'last_used_at': timestamp,
    'last_seq': 150000,
    'event_log': {
        'active_path': str(sessions / 'large-history.stream.ndjson'),
        'segment_count': 1, 'max_segment_bytes': 67108864, 'max_segments': 5,
    },
    'closed': True, 'title': None, 'messages': [], 'updated_at': timestamp,
    'cumulative_token_usage': {}, 'request_token_usage': {}, 'acpx': {},
}
(sessions / 'large-history.json').write_text(json.dumps(record))
with (sessions / 'large-history.stream.ndjson').open('w') as stream:
    for i in range(150000):
        stream.write(json.dumps({'jsonrpc': '2.0', 'id': i, 'result': {}}) + '\n')
PY
```

CLI command used for both builds:

```sh
HOME="$FIXTURE" node dist/cli.js --agent synthetic-agent \
  --cwd "$FIXTURE/workspace" sessions export large \
  --output "$FIXTURE/archive.json"
printf 'exit=%s\n' "$?"
```

Captured base result:

```text
Maximum call stack size exceeded
exit=1
```

Captured fixed result (temporary path normalized):

```text
exported session to <fixture>/archive.json
exit=0
```

The post-export check compares every ID with its original position:

```sh
python3 - <<'PY'
import json, os, pathlib
history = json.loads((pathlib.Path(os.environ['FIXTURE']) / 'archive.json').read_text())['history']
assert len(history) == 150000
assert all(event['id'] == i for i, event in enumerate(history))
print('Final built CLI: 150000 events preserved in exact order')
PY
```

Captured verification result:

```text
Final built CLI: 150000 events preserved in exact order
```

The committed regression additionally covers 300,000 distinct events across a rotated segment and the active segment.

</details>


steipete:
Landed in b67d67209096e924a64c8b010ff002fe3dddf416.

Large session exports now append events directly, removing the argument-limit crash while preserving the complete event order. The accompanying shutdown regression waits for an actual active prompt before signalling.

Validation passed: 1,069 local tests plus 152 coverage tests, docs, isolated P0–P2 review, and the built-CLI 150,000-event proof captured in the PR body. Exact head `6fea12b940d1201d97635d49c71a4f72fc6f60c4` passed every CI job, including Node 22/24/26, mutation, conformance, Slophammer, and CodeQL: https://github.com/openclaw/acpx/actions/runs/34936121981.

PR #580, opened 2026-09-15
fix: harden session storage permissions across rewrites

steipete:
## What Problem This Solves

Session-state rewrites could replace private files with permissions derived from the process umask. This change hardens the existing local storage defaults.

## User Impact

On POSIX systems, session records and indexes use `0600`, and the acpx session directory uses `0700`. Writes tighten older permissive files and the immediate session directory while preserving saved contents. This also covers the embedded runtime's file store.

This is hardening within acpx's trusted-local-user model. It does not provide isolation between agents sharing an OS account, alter session environment persistence, or add Windows ACL guarantees. Existing files are tightened when rewritten; this does not scan or delete historical records.

## Why This Change Was Made

Use one session writer backed by `@openclaw/fs-safe/temp` for private staging and cleanup. It resolves operator-managed directory aliases and preserves ordinary atomic rename publication without persistent locks or post-publication inode requirements. The dependency is `0.10.0`, within the repository's existing 48-hour release-age policy. File permissions are set on the temporary descriptor before atomic replacement rather than inherited from a potentially permissive destination.

## Evidence

- All six new permission cases failed on the original writers with `0664` instead of `0600` and pass after the change. They cover new records, private rewrites, legacy permissive records/directories, data reload, and indexed lookup.
- Added regression coverage for relocated session directories after demonstrating the initial strict-helper implementation rejected their symlinks. Concurrent-save coverage exercises a 220-character session ID and verifies staging cleanup.
- `pnpm run check` passed: formatting, types, lint, builds, 1,091 passing tests (two existing skips), and all 156 coverage tests above the required thresholds.
- `pnpm run check:docs` passed.
- One full rerun stalled in the replay-viewer server worker before any of that file’s tests started. That worker was terminated; its eight tests passed independently. The unchanged full gate was rerun to verify completion.
- Source-blind built CLI/public API proof passed private creation, permission restoration, retained conversation history, relocated session directories, and 80 concurrent saves from four separate processes with complete JSON and no staging leftovers. Final independent review is clean. PR CI passed on the pushed commit.

## Maintainer decision and captured behavior proof

Owner-only session storage is intentional and accepted for this hardening. The session directory, including the target of an operator-managed symlink, must be dedicated to the owning OS account. Writes remove group/other access from that directory; previously configured cross-user readers lose access. A directory owned by another account that cannot be chmodded is not a supported private-session target. This policy was approved as part of the requested hardening; it is not a new isolation boundary.

[CI on the pushed commit passed](https://github.com/openclaw/acpx/actions/runs/34999795794), including Node 22/24/26, mutation, conformance, Slophammer, docs, and CodeQL checks on the PR.

<details>
<summary>Commands, environment, and captured synthetic results</summary>

Local runtime proof used macOS, Node v24.21.0, umask `0002`, temporary HOME/workspace directories, and the compiled synthetic ACP adapter. No external provider or real credentials were used. Checkout and temporary absolute paths below are normalized to `$CHECKOUT` and `$PROOF`.

After `pnpm run check` built the CLI and fixture, each CLI command used these common arguments and the isolated home:

```sh
env HOME="$PROOF/home" node "$CHECKOUT/dist/cli.js" \
  --agent "node $CHECKOUT/dist-test/test/mock-agent.js --supports-load-session" \
  --cwd "$PROOF/workspace" --ttl 1 --timeout 15 --deny-all \
  sessions new --name synthetic-proof
```

Creation exited 0. The created session JSON and index were `0600`, and the sessions directory was `0700`. Both files were then deliberately chmodded to `0664` and the directory to `0775`.

Using the same common arguments, these actual CLI operations all exited 0:

```text
prompt --session synthetic-proof "echo synthetic-proof-alpha"
prompt --session synthetic-proof "echo synthetic-proof-beta"
sessions history synthetic-proof --limit 20
```

The first prompt's captured stdout showed a real reload and response:

```text
[client] initialize (running)

[client] session/load (running)
synthetic-proof-alpha

[done] end_turn
```

After that write, session JSON/index modes were restored to `0600`, and the directory to `0700`. The second prompt ran after queue shutdown and loaded the same session again. Final history retained all four entries:

```text
session: 955ca021-36f9-4531-a3de-d3dec1f09e3a (4/4 shown)
2026-09-15T17:03:16.464Z user      echo synthetic-proof-alpha
2026-09-15T17:03:16.464Z assistant synthetic-proof-alpha
2026-09-15T17:03:16.464Z user      echo synthetic-proof-beta
2026-09-15T17:03:16.464Z assistant synthetic-proof-beta
```

The CLI creation and public `createFileSessionStore({stateDir}).save(record)` operations were also repeated with their sessions directories symlinked to dedicated temporary `0775` targets. Both succeeded, preserved the symlinks, tightened targets to `0700`, and produced `0600` record/index files. A fresh public-store instance reloaded the saved record.

A synchronized four-process public-store probe made 20 distinct saves per process to one record ID. Captured results:

```json
{
  "outcome": "pass",
  "successfulSaves": 80,
  "finalTitle": "synthetic-title-0-19",
  "finalName": "worker-0",
  "finalSequence": 19,
  "completeRecord": true,
  "fileMode": "0600",
  "directoryMode": "0700",
  "directoryEntries": ["synthetic-concurrent-record.json"]
}
```

Each child recorded 20 successes, zero errors, and exit code 0. The final document matched one complete submitted record, and no staging/lock entries remained. This verifies ordinary last-writer behavior, not merged concurrent updates. Task-owned processes and state were cleaned afterward.

The committed, directly reproducible writer regressions use:

```sh
node --import tsx --test \
  --test-name-pattern='private.*permissive umask|concurrent saves' \
  test/session-persistence.test.ts test/runtime.test.ts
```

Captured result: 9 tests, 9 passes, 0 failures. These cover new/private/legacy/symlink cases at both actual storage entry points, plus long-ID concurrent saves and staging cleanup. Real FUSE/exFAT volumes and Windows ACL hardening were not tested or claimed.

</details>


steipete:
Landed as session-storage hardening in 3c83c9db1a71bcc0a310b9dc7fe942393d9ac065.

The PR body records the accepted owner-only directory policy and captured synthetic CLI/public-store proof, including symlink targets, conversation reloads, and 80 concurrent saves across four processes. Final independent review was clean; the full local gate, docs gate, and CI on the pushed commit passed: https://github.com/openclaw/acpx/actions/runs/34999795794

The release note remains under Unreleased. This does not change acpx's trusted-local-user boundary or environment persistence contract.

PR #581, opened 2026-09-15
fix: report missing files as resources not found

obviyus:
## What Problem This Solves

Reading a missing file returns a generic internal error to the agent.

## User Impact

Agents receive the standard ACP resource-not-found error and can distinguish missing files from other read failures.

## Why This Change Was Made

The existing filesystem request handler now converts only `ENOENT` to ACP error `-32002`. Permission and path failures keep their existing behavior.

## Evidence

- CLI wire regression verifies ACP error `-32002` and the file URI.
- Existing-file reads and outside-workspace rejection pass.
- `pnpm run check` and `pnpm run check:docs` pass.
- The real Gemini adapter receives the corrected error; its separate error-parsing limitation still blocks new-file creation through client filesystem requests.


obviyus:
Landed as b0971fa45cbf821229bc70c496bebf64df6c67b6.

Validation: full local code and documentation checks passed. [CI run 35007405139](https://github.com/openclaw/acpx/actions/runs/35007405139) passed at the reviewed head.

PR #582, opened 2026-09-15
feat: expose model selection to runtime clients

obviyus:
## What Problem This Solves

Embedding clients cannot select models through the public runtime when an adapter exposes only `session/set_model`.

## User Impact

Clients can call `runtime.setModel({ handle, model })` with the adapter's model ID. Accepted selections remain available after reconnect.

## Why This Change Was Made

The optional public method uses the existing session control owner, model dispatcher, and saved model state. It selects the adapter's advertised control and saves only accepted selections.

## Evidence

- Real stdio tests cover a custom model configuration key and legacy model controls through idle sessions, active turns, reconnect, and rejected changes.
- The actual Codex, Claude, Gemini, and Qwen adapters accept the public setter with supported localhost providers; their next requests use the selected model.
- `pnpm run check` and `pnpm run check:docs` pass.


obviyus:
Landed as 5da4ec0dc5a827df318820b320709ba6f6ef6d6c.

Validation: full local code and documentation checks passed. [CI run 35008566058](https://github.com/openclaw/acpx/actions/runs/35008566058) passed at the reviewed head. Public model selection also passed with the actual Codex, Claude, Gemini, and Qwen adapters using supported localhost providers.

PR #583, opened 2026-09-15
fix: contain ACP and replay-viewer file access

steipete:
## What Problem This Solves

ACP file operations could follow a symlink outside the selected working directory, and replay-viewer reads could block indefinitely on a named pipe.

## User Impact

This is filesystem hardening within acpx's trusted-user model. ACP and replay-viewer reads reject outside symlinks and special files. Contained aliases, large reads, and ordinary in-place writes continue to work; writes to hardlinked files are rejected. These checks do not provide an OS sandbox for terminal commands or hostile same-user processes.

## Why This Change Was Made

Use the existing fs-safe root owner for file admission and containment, removing the viewer's duplicate path resolution and descriptor checks. Viewer reads retain both the runs-directory and individual-bundle boundaries. ACP writes preserve file modes and inode identity, and missing-file reads retain the standard ACP resource-not-found response from #581.

Disable Vite's unused WebSocket listener: `hmr: false` still opens port 24678, while acpx already owns its live-update socket. CLI output tests run in child processes, avoiding interception of Node's test-runner transport. Failed HTTP binds close Vite and live-sync resources so startup failures terminate cleanly. Bound the full test suite to four concurrent files after measuring up to 5.9 seconds of event-loop starvation during otherwise 1–5 ms HTTP responses; assertions, timeouts, coverage thresholds, and the selected tests are unchanged.

## Evidence

- Baseline reproduction: ACP read and write through an outside symlink succeeded. The viewer FIFO regression timed out on the baseline and rejects on this branch.
- Focused filesystem/viewer tests preserve contained symlinks, hardlink reads, literal path names, executable mode and inode, reads above 16 MiB, and uniform HTTP 404 responses for denied/missing targets.
- Live proof used Node 24.21.0 on macOS, synthetic HOME/workspace files, the built CLI, and the compiled mock ACP adapter. Actual operations included:

```sh
node dist/cli.js --agent "node $REPO/dist-test/test/mock-agent.js" \
  --approve-all --cwd "$WORKSPACE" --format quiet exec "read $WORKSPACE/outside-file"
node dist/cli.js --agent "node $REPO/dist-test/test/mock-agent.js" \
  --approve-all --cwd "$WORKSPACE" --format quiet exec "write $WORKSPACE/outside-directory/new/nested.txt changed"
node dist/cli.js --agent "node $REPO/dist-test/test/mock-agent.js" \
  --approve-all --cwd "$WORKSPACE" --format quiet exec "write $WORKSPACE/contained-file shorter"
```

[...]

PR #584, opened 2026-09-15
fix: keep flow and session files private across writes

steipete:
## What Problem This Solves

Flow bundles, event logs, imported history, exports, and new config files could be readable by other local users under a permissive umask. Separate atomic writers also duplicated staging and cleanup logic.

## User Impact

This extends the private-file hardening in https://github.com/openclaw/acpx/pull/580. Owned state files and archives use POSIX `0600`; owned state directories use `0700`. Existing custom output and flow-source directories retain their permissions. Existing config files remain create-only, and append ordering and rotation limits are preserved. Maintainer decision: accept private modes for these artifacts and regular-file-only event logs and exports; output symlinks and user-selected directory permissions retain their documented behavior.

## Why This Change Was Made

Share one private-file writer backed by fs-safe temporary-file cleanup. Use private staging on the destination filesystem followed by ordinary rename, preserving overlapping-writer and filesystem compatibility without persistent lock sidecars. Flow and session logs use fs-safe regular-file admission before appending; no new size cap or durability policy is introduced.

Export follows user-selected output symlinks, including dangling links, using native path resolution so `symlink/../` and Windows cross-volume junctions retain their semantics. Existing special output files, including FIFOs, are rejected without replacement. This is hardening under the existing trusted-user model.

## Evidence

- The original flow/private-file permission regressions failed on the baseline and pass with the shared writer. A failed rename leaves no staging directory. A review-discovered FIFO replacement regression failed before its guard and now preserves the pipe and its alias.
- The live built CLI ran with Node 24.21.0 on macOS, synthetic homes, a permissive `0002` umask, and the compiled local mock ACP adapter with `--supports-load-session`. Representative actual commands, with common isolated-home/cwd/adapter options omitted:

```sh
acpx config init
acpx sessions new --name synthetic-source
acpx prompt --session synthetic-source 'echo synthetic-storage-alpha'
acpx sessions close synthetic-source
acpx sessions export synthetic-source --output "$OUTPUT/archive.json"
acpx sessions import "$OUTPUT/archive.json" --name synthetic-import --cwd "$IMPORT_CWD"
acpx prompt --session synthetic-import 'echo synthetic-storage-beta'
acpx sessions history synthetic-import
acpx flow run "$FLOW_SOURCE/proof.flow.ts" --input-json '{"value":7}'
```

Config initialization preserved existing customized bytes on its second call. Imported history retained the alpha turn and appended beta (four messages); source/imported records, indexes, and NDJSON were `0600`. A compute-only TypeScript flow imported a sibling function and persisted the expected value `42`. Run directories were `0700`, bundle/artifact/log files `0600`, source/output parents remained `0775`, and no loader scratch file remained.

Real CLI exports to existing and dangling symlinks and a literal `jump/../` path reached the intended targets while preserving their aliases and parent permissions. Exports to a FIFO and its alias failed promptly, preserving the pipe's type and inode.

Native Windows x64 (10.0.20348), Node 24.19.0, also passed four built-CLI export cases on a temporary AWS Crabbox host:

```text
node dist/cli.js --agent "node synthetic-acp-agent.mjs" --cwd <isolated-workdir> sessions export native-win-proof --output <output-alias>
```

Directory junctions to existing/missing targets and file symlinks to existing/dangling targets all exited 0, retained the alias, and produced the expected synthetic archive. The second drive used SUBST on the same volume; realpath mapped it back to C:, so this validates Windows alias compatibility rather than a distinct-volume before/after regression. The temporary state, mapping, and host were removed.

Validation on the rebased candidate: `pnpm run check` passed (1,107 tests: 1,105 passed, 2 platform skips; all 161 coverage tests passed the existing 85% gates). `pnpm run check:docs`, the 94-test focused storage/runtime suite, and the final isolated Codex review passed. The reviewed source and test blobs are unchanged by the changelog-only rebase onto #583.


steipete:
Landed as `e5c55570bc6203314b2a51978a9095a371f96059` from reviewed head `62b59a732b5df5750f45245607be8797eb50fbf1`. All CI checks passed, including Node 22/24/26, coverage, mutation, conformance, and CodeQL. Local full checks and live CLI storage/export/flow proof passed; native Windows alias exports also passed with the documented same-volume limitation. Main is synchronized and clean.

An additional combined-candidate live run with #583, #585, and #586 passed private storage, stdin startup without observed bootstrap payload files, FIFO, and same-provider recovery after SIGKILL.

PR #585, opened 2026-09-15
fix: pass queue startup options without temporary files

steipete:
## What Problem This Solves

Detached queue startup wrote credential-bearing options into a temporary file and relied on the child to remove it. Failed startup could leave that payload behind.

## User Impact

Queue startup now passes options directly over stdin. Large payloads and final startup diagnostics are preserved, while the temporary bootstrap file and its cleanup path are removed. Public CLI/session behavior is unchanged.

## Why This Change Was Made

The hidden queue-owner command is an internal self-spawn transport, so one pipe replaces the payload environment variables, tempfile producer/reader/cleanup, and duplicate spawn-option plumbing. Pipe errors are observed without discarding the child's final stderr. Malformed JSON errors omit the payload rather than quoting possible credential bytes.

This is hardening through simplification; it removes 81 net production lines and adds no dependency or public configuration.

## Evidence

- Focused owner parsing/spawn/lifecycle tests cover large Unicode payloads with a delayed reader, early pipe closure, final stderr diagnostics, warm owners, and shutdown. The new large-input regression failed on the baseline transport; the credential-redaction regression failed before its fix.
- Live proof used the built CLI with Node 24.21.0 on macOS, synthetic HOME/TMPDIR/config, and the compiled mock ACP adapter with `--supports-load-session`. A real terminal sleep kept the first prompt active while another submitter queued work:

```sh
acpx sessions new --name synthetic-queue
acpx prompt --session synthetic-queue 'terminal /bin/sleep 2'
acpx prompt --session synthetic-queue --no-wait 'echo synthetic-queued-proof'
acpx prompt --session synthetic-queue 'echo synthetic-warm-proof'
acpx status --session synthetic-queue
acpx prompt --session synthetic-queue 'echo synthetic-reconnect-proof'
acpx sessions close synthetic-queue
acpx sessions history synthetic-queue
```

Common isolated-home/cwd/adapter flags are omitted above. The queued call acknowledged before the active prompt completed, eight history messages retained order, a warm follow-up reused its owner PID, TTL expiry reported idle, reconnect created a new PID and loaded the existing provider session, and explicit close removed queue files.

Recursive filesystem-event observation plus 1,048 scans at 10 ms found no bootstrap payload filename or synthetic auth marker outside its intentional config. TMPDIR was empty afterward. This observation is bounded by event delivery/scanning; the deleted producer is also verified in source.

Validation: final isolated Codex review, docs, and all 65 focused owner/spawn/lifecycle tests pass. On the rebased candidate, the full check passed format/type/lint/build and 1,097 tests, with two platform skips and one viewer health timeout during three simultaneous local campaigns. The unchanged viewer suite then passed all nine tests in isolation, and all 160 coverage tests passed the existing 85% gates. The complete CI suite on the PR head must pass before merge.


steipete:
Landed as `e1b63a66f65aa7ba01e2ce10be88f4590d50aeb7` from reviewed head `377e748ffaa6615c92ed97ba1f6687e0eb892dd8`, after rebasing onto the landed private-storage change. All checks on that exact head passed, including Node 22/24/26, coverage, mutation, conformance, and CodeQL. The reviewed source/test blobs were unchanged by the rebase.

Live component and combined-stack proof preserved queue ordering, warm reuse, reconnect, and shutdown while observing no bootstrap payload file. The pipe transport removes 81 net production lines. Main is synchronized and clean.

PR #587, opened 2026-09-15
fix: preserve shared sessions and contain malformed I/O

steipete:
## What Problem This Solves

Flows and queued CLI prompts can both finish successfully while overwriting the same session's saved conversation. Separately, malformed viewer input can terminate the server, piped UTF-8 can be corrupted at read boundaries, and optional performance capture can write permissive files or block CLI exit on a FIFO.

## User Impact

Shared-session turns preserve both completions, including when the first turn lasts longer than 15 seconds. Waiting prompts can time out or cancel without changing the active conversation. Viewer input errors stay local to the request or connection. Unicode survives stdin chunk boundaries, and optional diagnostic captures use POSIX `0600` without changing existing parent-directory modes.

## Why This Change Was Made

Move the existing stream-lock responsibility to the shared session-turn boundary, before the authoritative history read and through final checkpoint/cleanup. The event writer no longer acquires a second, late lock. Live and same-process owners are preserved; dead-process recovery, incomplete reservation handling, owner-checked cleanup, and the existing lock-file format remain supported. Staged publication uses fs-safe cleanup and retains an exclusive-create fallback on filesystems without hardlinks.

One CLI input reader now serves prompt, exec, and compare, using Node's stream consumer to retain partial UTF-8 sequences. Performance capture uses fs-safe regular-file append. Maintainer decision: diagnostic symlink, hardlink, and special-file targets are intentionally skipped; capture remains optional and does not change command exit status. No new configuration, dependency, persistence format, or size limit is introduced.

The viewer validates message shape, handles socket/async failures, skips corrupt sort metadata, and removes redundant file-reader and decode wrappers. Valid protocol behavior, existing filesystem roots, aliases, and large reads remain supported.

This is scoped hardening and a concrete session-integrity repair within acpx's existing trusted-user model. The cancellable turn owner adds necessary lifecycle code; consolidation removes 92 production lines elsewhere. This follow-up is net +28 production lines, leaving the full hardening series net -11 relative to before [#583](https://github.com/openclaw/acpx/pull/583).

## Evidence

Before fixes on `92cd5f8`:

- Real flow/CLI overlap lost the flow completion both after a live lock was stolen at 16 seconds and with a short, correctly waiting contender. The first writer also removed its successor's lock in the long case.
- Malformed URL encoding, WebSocket JSON `null`, and invalid UTF-8 frames each changed a healthy viewer into a process exit 1 and connection refusal.
- Native stdin reads split inside `🦞` produced `����` in exec and both compare results.
- Metrics files were `0664` under umask `0002` and contained the synthetic prompt. A FIFO held an otherwise completed CLI until its proof timeout killed it.

Focused regressions fail on the original behavior and pass with the fixes. The independent live pass uses Node 24.21.0 on macOS, built CLI/server artifacts, temporary homes/workspaces, and the repository's compiled synthetic ACP adapter. Actual command forms include:

```sh
node dist/cli.js --agent "node $REPO/dist-test/test/mock-agent.js --supports-load-session" \
  --cwd "$WORKSPACE" --deny-all flow run "$FLOW"
node dist/cli.js --agent "node $REPO/dist-test/test/mock-agent.js --supports-load-session" \
  --cwd "$WORKSPACE" --deny-all prompt -s "$SESSION" 'stream-sleep 1500 CLI_SENTINEL'
ACPX_PERF_METRICS_FILE="$CAPTURE" node dist/cli.js --agent "node $REPO/dist-test/test/mock-agent.js" \
  --cwd "$WORKSPACE" --deny-all --format quiet exec 'echo SYNTHETIC_PRIVATE_PROMPT'
```

The flow holds its turn for 19 seconds. The CLI is submitted while it is active; wire ordering and saved history verify serialization and both completions. Separate timeout/interruption and SIGKILL probes verify waiting-client isolation and same-provider recovery. Stdin proof observes native pipe reads before releasing the remaining emoji bytes, rather than assuming separate writes produce separate reads.

Real HTTP/WebSocket proof verifies generic 404/protocol errors, only the malformed-frame client closing with code 1007, and a healthy peer continuing to receive pongs and a live patch from a new CLI-generated flow. Metric files stay `0600` on creation/append, parents remain `0775`, alias targets stay untouched, and FIFO output no longer holds the CLI open.

Validation: `pnpm run check` completed 1,141 tests (1,139 passed, two platform skips, zero failures) and all 161 coverage tests at the existing 85% thresholds. `pnpm run check:docs` passed. Slophammer found zero duplicate-code or dependency-boundary findings. The final combined isolated Codex review and ClawSweeper review found no actionable defects. The diagnostic compatibility choice is [explicitly accepted](https://github.com/openclaw/acpx/pull/587#issuecomment-5688410605). [CI passed](https://github.com/openclaw/acpx/actions/runs/35025543931) on reviewed head `eb880230d8b6d38a2e78745687b67ee850ba697b`, including Node 22/24/26 tests, coverage, mutation, build, format, typecheck, lint, conformance, Slophammer, and docs. CodeQL also passed.


steipete:
Maintainer decision: accept regular-file-only optional diagnostic capture for this PR. The documented restriction is intentional: captures can contain prompt data, and a FIFO must not hold an otherwise completed CLI open. Diagnostic pipelines should use a regular-file destination; symlinks, hardlinks, and special files are skipped while capture remains best effort.

The regression and live checks cover creation and append at 0600, unchanged parent permissions, untouched alias targets, and successful CLI exit with a FIFO destination. This resolves the compatibility decision raised in the [review](https://github.com/openclaw/acpx/pull/587#issuecomment-5688385158). No actionable code findings remain; landing will follow green CI on the reviewed head.

steipete:
Landed via squash merge after rebasing onto current main: [1e730db](https://github.com/openclaw/acpx/commit/1e730db59373a77473c6975da827628074b3b093). The merged tree matches reviewed and tested head `eb880230d8b6d38a2e78745687b67ee850ba697b`.

All [CI jobs](https://github.com/openclaw/acpx/actions/runs/35025543931) and CodeQL passed, including Node 22/24/26 tests, coverage, mutation, build, docs, conformance, and policy checks. Local validation completed with 1,139 passing tests, two platform skips, and 161 passing coverage tests. Both reviews found no actionable code defects; the diagnostic compatibility decision is recorded above.

Independent live proof exercised the built CLI and viewer with a synthetic ACP adapter: overlapping flow/CLI completions survive, waiting-client cancellation and killed-owner recovery work, malformed HTTP/WebSocket inputs leave healthy clients usable, split native stdin reads preserve Unicode, and private metrics capture skips aliases/FIFOs without blocking CLI exit.

The local main checkout is fast-forwarded and clean. This is landed hardening and session-integrity repair; no package release was performed.

PR #591, opened 2026-09-15
fix: reject false conformance passes and simplify validation

steipete:
## What Problem This Solves

The conformance runner can report a successful operation as a passing expected-error case: it catches its own unexpected-success assertion and validates it as the operation's error. Validation tooling also repeats checks and session fixtures already owned elsewhere.

## User Impact

Expected-error cases now fail when their operation succeeds, including empty expectations or broad message filters. The runner also clears its shutdown fallback timer after the adapter exits, so completed runs exit promptly. CI routing and persisted-key enforcement retain their existing behavior.

## Why This Change Was Made

Keep only operation failures inside expected-error matching, then reject unexpected success outside that catch. Preserve adapter termination deadlines and clear both shutdown timers when exit is observed.

Remove the unreachable changelog-only CI branch and duplicate serializer source/denylist checks, retaining actual serialized-key validation and required-key presence checks. Reuse existing session/temp-home fixtures and remove a test that only asserted fixture defaults. The patch removes 95 tooling lines and 68 test/support lines while adding the real runner regression and documenting expected-error semantics.

## Evidence

- The new runner-CLI regression failed on original code: exit 0 instead of expected exit 1. A separate live probe showed both expected-error cases incorrectly passing before and correctly failing after.
- The same two-case live probe took 2,955 ms before and 445 ms after, with the same adapter-operation timing and no stderr output; shutdown deadlines are unchanged.
- Extracted before/after CI scope logic matched for all 585 path sequences of length 0–3 across eight file categories, including empty, docs, changelog, workflow, code, tests, and mixed changes.
- 158 focused conformance, export/import, persistence, queue, package, and key-policy tests passed. The executable persisted-key lint passed.
- `pnpm run check` and `pnpm run check:docs` passed: 1,139 tests passed, two platform skips, zero failures, plus 161 coverage tests. Changed conformance/changelog Markdown also passed explicit formatting and lint.
- Fresh isolated Codex review found no actionable defects.

- The full 21-case conformance profile passed against the synthetic adapter, preserving read/write permissions, cancellation, structured prompts, and expected-error matching.

<details>
<summary>Runner commands and before/after output</summary>

The macOS / Node 24.21.0 probe wrote two temporary cases: each runs `new_session`, one with `expect_error:{}` and one with `expect_error:{"message_any":["error"]}`. A temporary profile selected both cases. It then executed:

```sh
node --import tsx conformance/runner/run.ts \
  --profile "$PROFILE" --cases-dir "$CASES_DIR" --cwd "$WORKSPACE" \
  --agent-command 'node --import tsx test/mock-agent.ts' --format json
```

Captured output excerpts:

```json
{"phase":"before","exitCode":0,"elapsedMs":2955,"stderr":"","totals":{"cases":2,"passed":2,"failed":0},"results":[{"id":"synthetic.expected-error.0","title":"synthetic.expected-error.0","passed":true,"durationMs":164},{"id":"synthetic.expected-error.1","title":"synthetic.expected-error.1","passed":true,"durationMs":127}]}
{"phase":"after","exitCode":1,"elapsedMs":445,"stderr":"","totals":{"cases":2,"passed":0,"failed":2},"results":[{"id":"synthetic.expected-error.0","title":"synthetic.expected-error.0","passed":false,"durationMs":159,"error":"session/new succeeded but error was expected"},{"id":"synthetic.expected-error.1","title":"synthetic.expected-error.1","passed":false,"durationMs":127,"error":"session/new succeeded but error was expected"}]}
```

The full profile also ran through the same runner without the custom profile/cases arguments:

```json
{"cases":21,"passed":21,"failed":0}
```

The synthetic README stayed unchanged and the runner removed its scratch write file. The separate CI-routing proof executed the extracted old and new shell classifiers over 585 path sequences and compared both output values.

</details>


steipete:
Landed as [093af43](https://github.com/openclaw/acpx/commit/093af435b53ff57536f9f8ebb7474a33d909abcc). Successful ACP operations now correctly fail cases that require an error, and completed adapter shutdown clears its fallback timer. Redundant validation and fixtures were removed without lowering coverage gates.

The before/after runner proof, full 21-case profile, source-blind behavior validation, full local checks, and independent review passed. Hosted CI is green on Node 22, 24, and 26, including coverage, mutation, docs, Slophammer, and CodeQL. Main is synchronized.

PR #593, opened 2026-09-16
refactor: simplify session persistence ownership and parsing

steipete:
## What Problem This Solves

Session persistence retains an unused process-closing path and repeated parsing, file naming, and index enumeration logic.

## User Impact

No user-visible behavior change. Stored records, private file permissions, session lookup, close, export/import, and pruning keep their current contracts.

## Why This Change Was Made

Remove the unused persistence-level close implementation so process termination remains with the existing queue-aware session lifecycle. Reuse existing session path helpers, inline trivial field setters, share string-map normalization and index file enumeration, and serialize directly without cloning the whole record. This removes 103 production lines.

## Evidence

A 1,104-case before/after comparison produced identical record acceptance, normalized values and own properties, serialized output, and canonical schema behavior. A built-CLI round trip covered new → prompt → list/show → close → export → prune → import → show → close/prune, preserving messages and provider identity with a new local record ID. Fresh isolated Codex review found no actionable defects.

`pnpm install --frozen-lockfile && pnpm run check` passed on AWS Crabbox `cbx_f33bb4cb0d1d` (Node 24.18.1, pnpm 11.26.0): 1,140 passing tests, two platform skips, and all 161 coverage tests above the unchanged 85% gates. Formatting, typechecks, lint, and CLI/viewer builds passed.

The first local full run hit a 15-second CLI deadline in the existing shared-session contention test; a targeted run then missed its five-second startup deadline while the workstation load average exceeded 100. An instrumented real CLI/flow probe subsequently observed correct lock handoff and both processes exiting in 5.9 seconds. The unchanged full suite passed on the isolated runner, including all three contention cases. No test timeout, assertion, or retry policy was changed.

Captured live round-trip result:

```json
{"passed":true,"messagesPreserved":true,"providerIdPreserved":true,"newLocalId":true,"indexEmptyAfterPrune":true}
```


steipete:
Landed as [b7fa2fe](https://github.com/openclaw/acpx/commit/b7fa2fe7be357ffcba17c1b973e39f642fb8ea1d), removing 103 production lines and leaving session closure with the existing queue-aware lifecycle.

The parser/serializer equivalence corpus, built-CLI persistence round trip, instrumented flow/CLI lock handoff, full AWS check, independent reviews, and all 21 exact-head CI checks passed. CI includes Node 22/24/26, coverage, mutation, Slophammer, and CodeQL. The local deadline misses and isolated proof are documented in the PR body. Main is synchronized and clean.

PR #602, opened 2026-09-16
fix: settle flow ACP capture writes before reporting results

steipete:
An ACP bundle append failure could terminate acpx as an unhandled rejection while the agent prompt was still running, leaving the saved flow marked `running`. Persistent and isolated steps duplicated capture arrays and only joined writes after successful prompts.

Give each prompt one capture owner that immediately handles append rejections, joins all pending writes after the ACP operation settles, and preserves the original agent error when both prompt and capture fail. Successful prompts surface capture failures as failed steps. FlowRunStore still owns sequence allocation and filesystem writes, and fs-safe still rejects changed or non-file destinations. Share the existing connection-option snapshot across the three ACP call sites.

Flow output is captured privately, so also use the existing error-emission policy to let the caller display adapter errors. Previously those errors could be marked as already shown. This is failure-handling hardening; permissions and filesystem restrictions are unchanged.

Production code shrinks by 14 lines. The process-level regression fixture adds 194 test lines to exercise real ACP transport and filesystem failure without production injection hooks.

Reproduction and validation:
- A synthetic child adapter waits until the outgoing prompt is recorded, replaces the manifest-listed `events.ndjson` file with a directory, emits an update, and holds its final response until released.
- All four strict child-CLI regression cases fail on the original code with early uncaught exits. They pass on the fix, which waits for release and persists a failed node/run.
- Built CLI before/after proof covers persistent/isolated × successful/failed adapter prompts. The candidate retains and displays `DISTINCTIVE_PROMPT_FAILURE` when the adapter fails; otherwise it reports the fs-safe append refusal.
- Normal built-CLI proof runs two persistent steps followed by one isolated step. It preserves shared/distinct session identities, captured text, and nonempty event ranges: `1–3`, `4–12`, and `1–9`.
- `pnpm install --frozen-lockfile && pnpm run check && pnpm run check:docs` passes on an isolated Linux runner. Local lint/typecheck, focused regressions, and independent review pass.

```sh
pnpm run build
pnpm run build:test
node --test dist-test/test/flows-capture.test.js
```

Captured behavior for each failure combination:

| Revision | Exits before adapter release | Saved run | Saved node | Raw uncaught error |
| --- | --- | --- | --- | --- |
| Before | yes | running | absent | yes |
| After | no | failed | failed | no |

The regression test drives the CLI with a disposable HOME and the committed synthetic adapter in `test/fixtures/flow-capture-agent.ts`. Outer flow timeout/cancellation ownership is unchanged.


steipete:
Landed as dd109c8c61514af32fe8d3fa3ae35bcf213c9251. Flow ACP capture now handles rejected appends immediately, joins pending writes on both prompt outcomes, preserves the agent failure when both fail, and reports captured errors through the caller. fs-safe restrictions are unchanged. Four strict child-CLI regressions fail on the original code and pass the fix; built fault and normal session/sequence proofs, full isolated code/docs checks, independent review and exact-head CI passed.

PR #604, opened 2026-09-16
refactor: prepare text tool output once

steipete:
Text tool updates summarized input, file locations, and output once for duplicate detection and again for rendering. Prepare those display strings once, then share the starting/final tool renderer. Keep the signature based on raw title/status/kind and untruncated output, preserve first-start and repeated-final suppression, and simplify string deduplication and assistant writes.

This removes 57 production lines. The existing lifecycle test now verifies that identical final updates stay suppressed and changed final output appears (+13 test lines).

Validation:
- All 22 formatter tests pass; local lint/typecheck and independent review pass.
- `pnpm install --frozen-lockfile && pnpm run check` passes on an isolated Linux runner.
- All 288 synthetic formatter cases are byte-identical before/after (271,464 UTF-8 bytes), independently rechecked from captured artifacts. Cases include ANSI/plain, read suppression, structured/circular/BigInt inputs, locations, repeated finals, changed text beyond truncation, partial updates, and explicit null clearing.
- Built CLI before/after proof uses a real synthetic ACP child emitting thought, assistant and tool lifecycle updates. An identical completed update prints once, a changed completed update prints again, and stdout/stderr remain byte-identical.

```sh
pnpm run build
node dist/cli.js --cwd "$PROOF_HOME" --agent "node text-render-agent.mjs" --deny-all exec 'render proof'
```

Captured CLI assertions:
```json
{"before":{"exit":0,"completedSections":2,"stdoutBytes":507},"after":{"exit":0,"completedSections":2,"stdoutBytes":507},"stdoutAndStderrIdentical":true}
```
No output format or public behavior changes are intended.


steipete:
Landed as f871d9c84252e239e12ca4bc1720f6000c32fcea. Tool display text is prepared once and shared across duplicate detection and rendering; existing status, suppression, spacing and truncation behavior is preserved. Production -57 lines, tests +13. All22formatter tests,288byte-identical scenario captures with independent artifact comparison, exact built-CLI before/after output, full isolated checks, independent review and exact-head CI passed.

PR #605, opened 2026-09-16
refactor: share prompt retry preparation

steipete:
Persistent prompts and temporary exec sessions repeated retry eligibility, backoff, and the check for output arriving during the delay. Share that preparation in one helper while retaining their different lifecycle cleanup and the persistent-session disconnect veto. Remove the private string marker used to continue the persistent loop.

This removes 19 production lines and 10 test lines overall. Consolidated integration cases preserve immediate partial-output coverage and add output delivered after the error, during backoff. The retry budget still belongs to each request, including calls through a warm queue owner.

Validation:
- All seven focused integration cases pass: strict-JSON recovery, both warm-owner budget directions, and immediate/late output vetoes in prompt and exec. The four output-veto cases also pass on the baseline.
- `pnpm install --frozen-lockfile && pnpm run check` passes on an isolated Linux runner. Local lint/typecheck and independent review pass.
- Built CLI proof exercises six cells against a real synthetic ACP adapter in fresh homes. Both modes recover from a pre-output error with two requests. Output before the error or during backoff leaves exactly one request and fails. Strict JSON emits no stderr in all cells.

Representative commands:
```sh
pnpm run build
node dist/cli.js --cwd "$PROOF_HOME" --agent "$SYNTHETIC_AGENT" --deny-all --format json --json-strict --prompt-retries 1 exec retryable-error-once
node dist/cli.js --cwd "$PROOF_HOME" --agent "$SYNTHETIC_AGENT" --deny-all --format json --json-strict --prompt-retries 1 exec late-retryable-error
```
The equivalent persistent cells create a named session and use `prompt -s proof`, then close it.

| Prompt scenario | Exec requests / exit | Persistent requests / exit | Strict JSON stderr |
| --- | --- | --- | --- |
| Retryable error before output | 2 / 0 | 2 / 0 | empty |
| Partial output before error | 1 / 1 | 1 / 1 | empty |
| Output after error during backoff | 1 / 1 | 1 / 1 | empty |

No retry defaults, delays, output policy, or public contracts change.


steipete:
Landed as 67699361271656817f73f2b70e581b15d44f52a1. Shared retry preparation owns eligibility, backoff and the fresh output check; persistent and exec cleanup remain separate. Production -19 lines and tests -10. Focused warm-owner and early/late-output tests, six built-CLI strict-JSON cases, full isolated checks, independent review and exact-head CI passed.

PR #606, opened 2026-09-16
refactor: centralize flow edge routing

steipete:
Successful and failed flow steps duplicated edge lookup, switch-value validation, and target selection. Use one internal resolver, with failure-edge eligibility checked before path parsing. Remove the separate outcome resolver and migrate its sole runtime caller.

Failed steps still follow only explicit `$result.*` switch edges. Direct edges, output switches, and invalid non-result paths leave the original execution error in charge. Eligible result switches retain their existing scalar and missing-case errors. Explicit node/output arguments remain independent of the publicly mutable result object.

Production code is 15 lines smaller. A 30-line test exercises the public FlowRunner boundary and verifies that failed direct/output/invalid edges preserve the exact original error and never launch the next step.

Validation:
- Five focused public-flow tests pass, including failure guards, timeout outcome routing, output/result publication, and falsy callback failures.
- All 348 before/after resolver scenarios produce identical targets or error messages, independently checked from captured artifacts.
- Built CLI proof runs a five-step flow twice with different inputs: compute → function action → shell → rejected compute → checkpoint. The explicit `$result.outcome` edge reaches waiting state; receipts and persisted output match the CLI.
- `pnpm install --frozen-lockfile && pnpm run check` passes on an isolated Linux runner. Local lint/typecheck and independent review pass.

Representative live commands:
```sh
pnpm run build
node dist/cli.js --cwd "$PROOF_HOME" --deny-all --format json flow run completion.flow.mjs --input-json '{"value":"FIRST"}'
node dist/cli.js --cwd "$PROOF_HOME" --deny-all --format json flow run completion.flow.mjs --input-json '{"value":"CHANGED"}'
```

Both captures report:
```json
{"status":"waiting","outcomes":["ok","ok","ok","failed","ok"],"functionReceipt":true,"shellArtifacts":true,"diskMatchesCLI":true}
```
The removed helper was internal to the bundle; no public API, flow schema, or output format changes.


steipete:
Landed as d0137d66dbac6206926cc26f2171e099605a77d0. One internal resolver now owns success/failure edge selection, retaining explicit failure-routing eligibility and original error behavior. Production -15 lines. Public FlowRunner failure guards,348identical resolver results, two built-CLI runs, full isolated checks, independent review and exact-head CI passed.

PR #608, opened 2026-09-16
fix: prefer non-aborting Codex permission refusals

### LINKED ISSUE #535: Refusing a permission request can abort the whole turn: pickOption treats all reject_once options alike
## Summary

`pickOption(options, ["reject_once", "reject_always"])` in `src/permissions.ts` matches on `kind` alone and returns the first hit. codex-acp offers **two** options with kind `reject_once` whose consequences differ enormously, and acpx cannot tell them apart:

| optionId | label | codex decision | effect |
|---|---|---|---|
| `decline` | "No, continue without running it" | `Denied` | the tool call fails, **the turn continues** |
| `cancel` | "No, and tell Codex what to do differently" | `Abort` | `sess.interrupt_task()` — **the whole turn is interrupted** |

So whenever codex happens to list `cancel` first — or offers it as the *only* refusal — acpx answers a "no" with "abort everything".

## Why this is not a rare edge case

codex-acp's `defaultCommandDecisions` (v1.7.0) has:

```js
if (params.additionalPermissions) return ["accept", "cancel"];
```

Sandbox-escalation approvals — the prompt raised when the model retries a command with `sandbox_permissions: "require_escalated"` after a sandbox/network failure — take that branch. `decline` is **not offered at all**. The only way acpx can say no is the option that kills the turn.

Escalations are common in sandboxed runs because codex's own `on_request` prompt template instructs the model to retry with `require_escalated` whenever a command fails with a likely sandbox network error (DNS, host resolution, registry access). A long agent turn can therefore be destroyed by its first refusal, after tens of thousands of tokens of work, with nothing recoverable — the in-flight tool comes back as `aborted by user after 0.1s`, the turn ends `TurnAborted { reason: Interrupted }`, and `session/prompt` returns `stopReason: "cancelled"`.

## Every refusal path is affected

All of these funnel through the same `rejectOption` computed at `src/permissions.ts:380`:

- `mode: "deny-all"` → `resolveModeMatch` → `selectedOrCancelled(rejectOption)`
- `--approve-reads` (or default) on a non-`read`/`search` tool with no TTY → `resolveNonInteractivePermission` → `selectedOrCancelled(rejectOption)`
- policy `action: "deny"` → `resolvePolicyMatch` → `selectedOrCancelled(rejectOption)`
- policy `action: "escalate"` with no TTY → `resolveEscalatingPermissionRequest`
- **the interactive prompt too**: answering `N` at `[permission] Allow …? (y/N)` reaches `resolveInteractivePromptResult` (`src/permissions.ts:272`) → `selected(rejectOption.optionId)`

That last one is the surprising part: a human at a terminal declining a single escalation ends their whole session, with no indication that "no" meant "stop".

`decisionToResponse` (`src/permissions.ts:419`) has the same problem — `DECISION_FALLBACK_ORDER` maps `reject_once` to a kind lookup, so a programmatic `{ outcome: "reject_once" }` decision can also land on `cancel`.

## Downstream chain, for reference

- codex-acp `selectedDecision()` maps the chosen optionId back to `"cancel"`
- codex `app-server/src/bespoke_event_handling.rs`: `CommandExecutionApprovalDecision::Cancel => ReviewDecision::Abort`
- codex `core/src/session/handlers.rs`: `ReviewDecision::Abort => sess.interrupt_task()`
- codex `core/src/tools/parallel.rs`: the running tool is aborted and reports `aborted by user after {secs}s`

## A second, related issue: `cancelled()` as the deny fallback

`selectedOrCancelled` and `resolveNonInteractivePermission` fall back to `cancelled()` when no reject option is found, and `resolvePermissionRequestWithDetails` returns `cancelled()` for an empty options array. `outcome: "cancelled"` is the *most* destructive answer available — codex-acp maps it to `decision: "cancel"` as well. Using it as the fallback for "I wanted to deny but found nothing to select" turns a missing option into an aborted turn, silently.

## Why the tests don't catch it

Every permission fixture offers exactly one `reject_once` option:

- `test/permissions.test.ts` — `{ optionId: "reject", kind: "reject_once" }`
- `test/mock-agent.ts:1406`, `test/compare-command.test.ts:146,160` — same
- `conformance/runner/run.ts:229` picks by `option.kind === "reject_once" || option.kind === "reject_always"`

The two-competing-`reject_once` shape that codex actually sends never appears in a fixture.

## Repro

Any codex run whose sandbox blocks network, where the model retries a network command with `require_escalated`, under `--deny-all` (or non-interactive with the default deny policy, or interactively answering `N`). The turn ends immediately with `stopReason: "cancelled"` instead of the command being declined.

Unit-level equivalent — a request shaped like a codex escalation prompt:

```ts
const request = {
  sessionId: "s",
  toolCall: { toolCallId: "1", kind: "execute", status: "pending" },
  options: [
    { optionId: "allow_once", name: "Yes, proceed", kind: "allow_once" },
    { optionId: "cancel", name: "No, and tell Codex what to do differently", kind: "reject_once" },
  ],
};
await resolvePermissionRequest(request, "deny-all");
// → { outcome: { outcome: "selected", optionId: "cancel" } }  ← aborts the turn
```

## Suggested fix

1. **Rank refusals by optionId, not only by kind.** Prefer known-safe ids (`decline`, `reject`, `reject_permissions`) over abortive ones (`cancel`, `abort`); use kind as the tiebreaker. Since ids are adapter-specific, a small per-agent table (`agents/`) would be a more honest home than a global regex.
2. **Never silently pick an abortive refusal.** When the only refusal available ends the turn, surface it — acpx already has the right shape for this in the `permission_escalation` event — so callers can distinguish "the agent declined a command" from "the turn was destroyed by a decline".
3. **Stop using `cancelled()` as the deny fallback.** Reserve it for an actual cancel; if no refusal option exists at all, report that rather than quietly escalating to abort.
4. **Add a fixture with two `reject_once` options** (`decline` + `cancel`) and assert the safe one is chosen, in both the unit tests and the conformance runner.

Upstream, the cleanest fix is for codex-acp to re-attach the decision to each option's `_meta` — earlier versions did (`_meta: { codex: { decision: "decline" | "cancel" } }`), and 1.7.0 dropped it, leaving optionId as the only discriminator on the wire. Worth raising there too, but acpx can behave correctly today without it.

Longer term this is arguably an ACP protocol gap: `reject_once` / `reject_always` encode *recurrence* but say nothing about *blast radius*, so every client has to guess which refusals end the turn.

Versions: acpx `v0.13.2-6-g95a9466`, codex-acp `1.7.0`.

steipete:
## What Problem This Solves

Codex can offer both `decline` and `cancel` as `reject_once` permission choices, although `cancel` can interrupt the current turn. acpx previously selected whichever appeared first, so declining one operation could silently stop the turn.

## User Impact

For the verified Codex ACP adapter, prefer an offered `decline` or `reject_permissions` choice. Preserve approval priorities, the host callback's original request, and opaque option IDs from other adapters. If the adapter offers only cancellation, keep the safe refusal and explain its possible effect in text output, quiet stderr, JSON response metadata, and embedded runtime status. Explicit caller/session cancellation remains unchanged. Diagnostic subscriber failures cannot change the resolved permission decision. No operation is approved merely to keep a turn running.

## Why This Change Was Made

The compatibility rule lives at the Codex adapter boundary, where option IDs have verified meanings. Generic ACP permission selection retains its existing kind-based contract.

## Evidence

- The original implementation failed four permission-routing/diagnostic regressions; all 136 focused client, permission, and formatter tests pass, including throwing diagnostic observers.
- Independent public CLI/runtime validation passed 79 assertions across 25 synthetic scenarios, including interactive denial and both refusal orderings.
- A real Codex ACP 1.12.0 request offered only `cancel` as refusal. acpx denied it, emitted the notice, and returned `cancelled`; the attempted external task-owned write did not happen. Preference for a non-cancelling option is proven by protocol fixtures because that option was not offered in the live probe.
- Full Linux `pnpm run check` passed: 1,186 tests passed with 2 platform skips, plus all 171 coverage-target tests. Docs, skill, and Codex guide checks passed. [Linux run](https://crabbox.openclaw.ai/portal/runs/run_38bc710e605c).
- Built-client fault injection with global `approve-all`, a host rejection, and a throwing notice subscriber still sends `cancel` with notice metadata and leaves the attempted sentinel absent. The explicit approval control sends `allow` and creates its sentinel. Mode-based denial also remains safe under the same observer fault.
- Independent review and exact-head GitHub CI are rechecked for the final candidate before landing.

Fixes #535. Thanks @odrobnik for the detailed report and reproduction.


steipete:
The observer-failure finding is reproduced and fixed. The permission decision now survives a throwing diagnostic subscriber; its cancellation response and `_meta.acpx.permissionNotice` remain intact.

The regression tests failed on the first PR revision: the host-denial case returned `allow`, and mode-based denial returned an RPC error. Both now return the expected refusal, with one recorded permission decision.

An independent probe exercised the built `AcpClient` against a synthetic ACP agent that performs a real task-owned write only after an approval response:

| Case | Actual ACP option | Stop reason | Sentinel |
| --- | --- | --- | --- |
| Before repair: global approve-all, host reject_once, throwing notice subscriber | allow | end_turn | Created |
| After repair: same fault injection | cancel | cancelled | Absent |
| After repair: ordinary host rejection | cancel | cancelled | Absent |
| After repair: mode deny-all with throwing subscriber | cancel | cancelled | Absent |
| After repair: explicit approval control | allow | end_turn | Created |

The fault remained active in both repaired denial cases: the notice subscriber threw exactly once. Both actual refusal responses retained their notice metadata. Owned processes, state, and sentinels were cleaned up. Independent review of the repaired candidate found no actionable issues.

steipete:
Landed in 7416cff07130e97a465d2edc0d1502ff09bf365a.

Codex refusals now prefer an offered non-aborting choice. When only cancellation is available, acpx preserves the safe refusal and explains its effect. Diagnostic observer failures cannot replace a completed denial with approval.

Validation: full Linux check/docs passed (1,186 tests plus 171 coverage-target tests), 136 focused tests passed, and independent live/protocol proof covered normal refusal, a real Codex cancel-only request, and throwing-observer denial with actual sentinel controls. [CI passed on the exact landed PR head](https://github.com/openclaw/acpx/actions/runs/35084876786); the [final public review](https://github.com/openclaw/acpx/pull/608#issuecomment-5695842977) has no actionable findings. The fault-injection evidence is [above](https://github.com/openclaw/acpx/pull/608#issuecomment-5695960057).

Thanks @odrobnik for the detailed report.

Issue #528, opened 2026-08-29
Add passive session event watching


mavam:
Event streams currently belong to the client that submitted the prompt. `sessions read` exposes persisted history, but another process cannot follow an active session. This makes live debugging and monitoring difficult.

I propose a replayable watch API and matching CLI command:

```ts
runtime.watchSession({ handle, cursor, signal });
```

```console
acpx codex sessions watch -s reviewer
```

A watcher should receive replayed and live events with a monotonic cursor, request identifier, and terminal turn status. Attaching or disconnecting must not affect the turn.

This relates to #526 but would also benefit CLI-only users. I am happy to contribute it as a PR if this direction fits the project.

steipete:
Maintainer decision: approved. Implement passive observation for CLI/shared sessions with retained replay, exclusive opaque cursors, request IDs and settled turn outcomes. Watcher cancellation, disconnects and broken output pipes must leave the turn and owner lifetime alone. Owner loss must remain an unknown outcome instead of success or an automatic retry.

The chosen format adds explicitly local segment/turn records to the existing journal while preserving actual ACP messages. This is an intentional on-disk extension: direct file consumers must distinguish local records from entries with `jsonrpc: "2.0"`; built-in readers, ACP-only prompt stdout and portable archives remain compatible. The documented Breaking entry accepts that tradeoff to obtain correct request attribution, cursor recovery and terminal outcomes without introducing a second event store. Legacy/imported history keeps its actual contents and does not gain fabricated local metadata.

The implementation has passing Linux code/docs checks, independent behavior validation including a 96 MiB bounded-memory replay, and an authenticated native proof with simultaneous API/CLI observers. Atomic replacement in #527 remains separate and deferred. Thanks @mavam for the proposal.

PR #614, opened 2026-09-16
feat(runtime): share local session owners with CLI clients

### LINKED ISSUE #526: Expose queue-backed sessions through acpx/runtime
`AcpRuntime` manages ACP connections in-process, while the CLI uses a separate queue owner and IPC path. An embedded application therefore cannot start a session that external `acpx` processes can join and control through the same live connection.

I propose exposing a queue-backed session client or adding a broker-backed mode to `AcpRuntime`. The existing API could remain largely unchanged:

```ts
const session = await runtime.ensureSession(...);
const turn = runtime.startTurn(...);
await runtime.cancel(...);
```

These operations would use the same session lookup and queue owner as the CLI. Exactly one adapter connection should own each session, regardless of which client starts it.

I am happy to contribute this as a PR if this direction fits the project.

steipete:
Applications can now create a named persistent session through `createSharedAcpRuntime()` and let separate CLI processes join, enqueue prompts, or cancel work through the same live adapter connection. Fixes #526. Thanks @mavam for the proposal.

The shared runtime uses the canonical CLI session store and queue owner. Queue and execution code now lives under `src/session`, with all CLI/runtime callers migrated and the old paths removed. Common turn-stream helpers are shared with the existing in-process runtime, which retains its custom storage and callback contracts. The new guide documents the deliberately narrower shared options and exact session lookup.

The owner preserves caller request IDs, reports actual ACP prompt start separately from queue acceptance, rejects duplicate active/queued IDs, and supports cancellation of one queued/starting/active request without cancelling another client's turn. Concurrent API/CLI ensures choose one named record. Client shutdown detaches and settles local operations while accepted work continues; explicit session close retains CLI soft-close semantics.

Shared submission and targeted cancellation require an advertised owner capability. Older owners fail clearly before receiving these new operations. Explicit session-wide cancel/close continue to mean the same thing as their existing CLI counterparts. Connection failures after writing a request report a non-retryable uncertain outcome; they do not kill the owner and replay a potentially executed prompt. Shared persistent turns preserve the saved provider session and reject failed resume instead of silently starting another conversation.

Validation:

- Independent source-blind validation passed both join directions, identical CLI/API session IDs and one live adapter PID, FIFO/history, concurrent ensure, targeted cancellation, duplicate IDs, observer detachment, legacy-owner safety, replacement-owner safety, crash uncertainty/no replay, diagnostics, and the existing direct runtime as a control. All 23 fixture session records were closed and owned processes cleaned.
- Real signed Junie 26.09.3 (3218.3), authenticated through its OpenAI BYOK mode, returned `SHARED_API_59147` through the built shared API and `SHARED_CLI_82604` through a separate CLI invocation. Both returned `end_turn`; the second turn reused the same native process. Total 32.34 seconds, exit 0. Fresh HOME/XDG/JVM state, default config/MCP/skill discovery disabled, and explicit task-local Keychain/browser denial were used; this does not claim interactive account-login coverage. Native processes and state were cleaned.
- The broad focused run passed 253 tests and exposed one new fixture missing its load capability; correcting that fixture made the actual-owner test pass. The final focused API/IPC/parser/process suite passed all 60 tests.
- Full Linux `pnpm run check` and `pnpm run check:docs` passed: 1,206 tests passed, two platform skips, plus all 171 tests in the configured coverage lane. AWS Crabbox run `run_77026380aaecbef32934925c5343332a`; lease `cbx_fa1dd46c42d9` is released with cleanup complete.
- Isolated Codex review of the complete candidate is clean through P2. The bundled older helper could not handle the relocation diff size; the maintained helper reviewed the entire frozen candidate with isolation and secret scanning intact. Initial Linux lint findings were fixed and the full gate rerun successfully.

The production change is net +543 lines for the new public client, targeted control, compatibility checks, and error semantics. Existing owners were moved and reused, with no parallel broker or storage implementation. Passive watching is being handled separately in #528.


steipete:
Maintainer acceptance of the compatibility tradeoff: retain the documented uncertainty contract, as recommended by the review.

This applies deliberately to existing CLI submissions as well as the new shared client. Once request bytes have been written, a malformed or lost response does not establish that the prompt was unexecuted. Automatically replacing the owner and replaying that prompt could repeat write-capable work. Such failures now preserve the owner and report non-retryable uncertainty; callers inspect/reconcile session state before choosing new work. Connection failure before a request is written retains normal startup/reconnect handling.

This decision is within the approved lifecycle/recovery scope in https://github.com/openclaw/acpx/issues/526#issuecomment-5700963014. The shared-session guide and PR body document it. Focused tests and independent process-level probes prove no replay and preservation of original/replacement owners. Full Linux validation and all exact-head CI jobs passed at b4aa3ee6226ac45fe2add8f3167df3286d4ae98e. There are no outstanding code findings; the identified merge-risk decision is resolved.

steipete:
Landed in 248676c605d81e32be1631dd199eb34d3c33cc72. Full Linux checks/docs, exact-head CI on Node 22/24/26, mutation checks, independent review, and source-blind API/CLI validation passed. Real authenticated Junie turns through the API and CLI reused one live adapter process. The intentional no-replay compatibility tradeoff is accepted in https://github.com/openclaw/acpx/pull/614#issuecomment-5701082020.

#526 is closed and canonical main is synchronized and clean. Passive watching continues separately in #528. Thanks @mavam for the proposal.

PR #621, opened 2026-09-19
fix(runtime): bound initialization and retire abandoned launches

saariuslystoned:
Embedded session creation could wait forever when the configured agent never answered ACP initialization, leaving the agent and its descendants running after each retry. Apply the existing positive `timeoutMs` to initial `client.start()` and let the runtime owner close that client before rejecting with `ACP_SESSION_INIT_FAILED`.

A launch that completes after `close()` now retires only its own child. It cannot adopt the stale process or close a replacement connection. Spawn/exit lifecycle observations are preserved, and the diagnostic names only the executable so configured arguments do not leak.

The deadline covers initialization; cleanup may extend beyond it. Unset/nonpositive timeouts retain existing unlimited behavior. Host admission callbacks retain their documented cancellation/timeout responsibility, and later `session/new`/`session/load` requests are outside this change.

Regression coverage uses real processes for silent initialization, repeated failures, descendant cleanup, unrelated sibling survival, closure before spawn, and replacement-connection reuse after an abandoned launch. Documentation and the Unreleased changelog describe the exact boundary.

Related: https://github.com/openclaw/openclaw/issues/147352. Thanks @saariuslystoned for the original fix and reproduction.


steipete:
Landed as 092c40c7cacfe8fe5179eee45bf1280ab4783559 after rebasing onto current main and tightening abandoned-launch ownership. Thanks @saariuslystoned!

Validation on the landed change:
- macOS focused process regressions: `pnpm run build:test` and `node --test dist-test/test/runtime-startup-timeout.test.js dist-test/test/client-process-cleanup.test.js` — 20 passed.
- Linux Node 24: `pnpm run check` — 1,282 passed, 2 skipped; declared flows/runtime coverage gates passed (93.71% lines/statements, 88.90% branches, 97.12% functions). `pnpm run check:docs` also passed in the Git worktree.
- Independent packaged-runtime comparison: published 0.17.0 remained pending at 2.3 seconds with a 500 ms timeout; this change rejected at 744/740 ms across two attempts, after stopping both agent and descendant while preserving an unrelated sibling. Healthy startup, unlimited defaults, redacted arguments and late admission also passed.
- Independent held-request probe: an abandoned child exited while the replacement had `session/new` pending; replacement lifecycle stayed unchanged and its delayed request completed. This confirms the existing current-child exit guard and rejects the branch review's sole advisory finding. No accepted/actionable findings remain.
- Final-head CI: https://github.com/openclaw/acpx/actions/runs/35486921691 — all 12 CI jobs passed, including Node 22/24/26, mutation, conformance, Slophammer and docs.

The deadline covers initialization, with cleanup allowed to extend it. Host admission callbacks retain their documented responsibility; later session creation/load deadlines and abrupt host death are outside this fix.

PR #628, opened 2026-09-20
fix: align bootstrap config selection and error output

steipete:
Repeating top-level `--cwd` currently loads configuration from the first directory while Commander executes in the last. An approve-all first project can therefore authorize writes inside a final deny-all project. Repeated `--mcp-config` has the same first-value mismatch, and configuration errors occur outside the normal output formatter.

Use one shared last-value scan for bootstrap paths, preserving required option values, command boundaries, empty cwd, and literal whitespace. Resolve the final MCP path against the final cwd. Move configuration loading into the existing error boundary with an explicit-output fallback before configuration becomes available; loaded format defaults still govern later errors. This removes the redundant first-match helpers without adding flags or changing exit/error codes.

Regression command: `pnpm run build:test` followed by `node --test --test-name-pattern='^bootstrap' dist-test/test/cli.test.js`. Before the fix, 24 cases failed and the version/default-format control passed. A synthetic `--cwd FIRST --cwd=SECOND --format json fixture exec 'write SECOND/sentinel.txt denied'` returned exit 0 despite SECOND's deny-all policy. After the fix it exits 5 and the file is absent. The focused bootstrap/version/config/output selection now passes all 31 tests.

Separate source-blind validation passed 51 cases over five contracts using the built CLI and a self-authored ACP peer. Actual `session/new.cwd` and peer cwd matched the final project; a positive approve-all control wrote its synthetic file, while deny-all returned `Permission denied for fs/write_text_file` and created none. Final relative MCP configuration reached `session/new` as:

```json
{"name":"second-tools","command":"synthetic-mcp","args":["second"],"env":[]}
```

Malformed global/project JSON, invalid timeout, and missing MCP files each exited 1. JSON/strict modes emitted exactly one JSON-RPC error with `id:null`, `error.data.acpxCode:"RUNTIME"`, and `origin:"cli"`, with empty stderr. Quiet emitted one stderr diagnostic; text emitted no stack trace. Boundary, literal-option-value, empty/whitespace-cwd, malformed-earlier-project, version-bypass, and no-agent-launch default-format probes also passed. All synthetic peers exited and fixtures were removed.

`pnpm run check` and `pnpm run check:docs` passed on AWS Crabbox `cbx_232ff861a01d`, run `run_3245442fa92e121de3a376fb44a17ab5`: 1,316 tests passed, 2 skipped, and the declared flows/runtime coverage gates passed. Isolated autoreview through P2 is scoped-clean. Config/output docs and Unreleased changelog updated.


steipete:
Landed as 5dc546b7bc6b2ee927d3e89272810a3146118fc3. Startup configuration now follows the same final cwd/MCP values as command execution, and configuration errors honor explicit JSON/quiet formatting.

The regression suite reproduced the wrong-project permission decision before the fix. Afterward, 31 focused tests, 51 source-blind CLI/ACP cases, and full checks with 1,316 passed/2 skipped were green; declared coverage, docs, and independent review through P2 passed. All 12 CI jobs passed for 512496f8882dd49f604a52e840c1f1f54e5389d6: https://github.com/openclaw/acpx/actions/runs/35498194369. Main is synchronized and clean.

PR #629, opened 2026-09-20
fix: serialize interactive permission questions

steipete:
Concurrent permission requests currently create separate readline interfaces over the same stdin. One `y` therefore approves multiple pending requests. Serialize tool, file-write, and terminal questions through their existing shared prompt helper, keeping each question's header/details with its input owner.

The queue retains ownership until the actual question finishes, even when an optional helper signal cancels a waiting caller. Active cancellation closes readline, errors release the slot, and EOF denies active/waiting questions without reopening ended input. A question-local controller disposes native readline listeners on close. Full ACP turn/request cancellation propagation is a separate follow-up; this PR fixes input ownership and provides the helper's cancellation support.

The native-readline child regression (`pnpm run build:test`, then `node --test --test-name-pattern='concurrent permission prompts' dist-test/test/permission-prompt.test.js`) sent `y` for the first question and `n` for the second. Before the fix it returned `[true,true]`; afterward it returns `[true,false]`. All 74 focused permission/helper/filesystem/terminal tests passed, including buffered `y\ny\n`, active/queued abort, failed questions, EOF, and abort-listener cleanup.

Source-blind validation exercised the built CLI through a synthetic ACP peer and a TTY-only preloader; native readline consumed real stdin. All four behavior contracts passed (85 checks). Both edit requests were issued before any response. Before the fresh second answer, the wire contained only:

```json
{"jsonrpc":"2.0","id":100,"result":{"outcome":{"outcome":"selected","optionId":"allow"}}}
```

After a separate `n`, the second response selected `deny`. Extra lines delivered in the first input chunk did not approve the second request. Mixed tool/file/terminal requests stayed sequential; only the approved write appeared, while the denied terminal marker remained absent. A positive terminal control produced its marker. EOF denied both requests without displaying the waiting header, a fresh invocation worked, and noninteractive/approve-all controls retained their behavior. Synthetic state and child processes were cleaned up.

`pnpm run check` and `pnpm run check:docs` passed on AWS Crabbox `cbx_232ff861a01d`, run `run_aea5d4914eb826731d75ec7fb5075172`: 1,297 passed, 2 skipped, with the declared flows/runtime coverage gates. Those full checks ran on parent 95c9d79; the branch was then fast-forwarded onto the independently validated bootstrap fix, preserving this PR's source changes. Final combined-head CI is required before merge. Isolated autoreview through P2 is scoped-clean. Permission docs, usage skill, and Unreleased changelog updated.


steipete:
Landed as 9b73767d155fbf57cf8aa2ba4ba51b8a8752f8cb. Tool, file-write, and terminal permission questions now share one input owner, so a single answer cannot approve multiple requests; EOF also releases the active question and denies queued ones.

Proof includes the native-readline before/after regression ([true,true] → [true,false]), 74 focused tests, full Linux checks and coverage/docs gates, and 10 source-blind built-CLI cases with 85 assertions covering actual mixed-operation effects. Independent review through P2 is scoped-clean. All 12 combined-head CI jobs passed for 70d891644700cf39bd32439d279b50f55ad89055: https://github.com/openclaw/acpx/actions/runs/35498625047. Main is clean and synchronized. Full ACP cancellation-lifetime propagation remains a separate follow-up.

PR #630, opened 2026-09-20
feat: add --skills-dir and --additional-dir flags

wtfsayo:
## Summary

Adds two repeatable global flags that feed ACP `additionalDirectories`:

- `--skills-dir <dir>` — points at a directory of skill folders (`<dir>/<name>/SKILL.md`). For each dir, acpx materializes a synthetic workspace root under `~/.acpx/skills-roots/` exposing it as both `.claude/skills/` and `.agents/skills/`, then sends that root on `session/new`, `session/load`, and `session/resume`.
- `--additional-dir <dir>` — sends a resolved workspace root verbatim through the same field, without the synthetic-root wrapping.

## Behavior

- Session creation fails (`AdditionalDirectoriesUnsupportedError`, USAGE) when the agent does not advertise `sessionCapabilities.additionalDirectories`.
- Reconnects (`session/load`, `session/resume`, and the load→fresh-session fallback) drop persisted dirs with a warning instead of failing, so a downgraded agent or deleted skills dir cannot wedge a session.
- Both flags resolve relative paths from `--cwd` and persist on the session record for reconnects (normalized to absolute against the session cwd).
- Agents that delegate fs to the client get read/write access to the advertised roots via `FileSystemHandlers.setAdditionalRoots`; paths outside every granted root remain blocked.
- Queue IPC, owner merge, WSL `wslpath` translation, and `--json-strict`/`--quiet` stderr gating all handled.

## Real behavior proof

```
$ acpx --skills-dir /tmp/acpx-skills-demo codex exec 'use the acpx-demo-skill'
[client] session/new (running)
[tool] Read file '/private/tmp/acpx-skills-demo/r7/myskill/SKILL.md' (completed)
  output: { "formatted_output": "---\nname: acpx-demo-skill\n..." }
```

Codex (codex-acp) advertised `sessionCapabilities.additionalDirectories`, received the synthetic root, discovered `acpx-demo-skill` inside it, and read its `SKILL.md` through the symlinked layout — end-to-end delivery confirmed on a real adapter.

## Synthetic roots

Shared roots are keyed by sha256-16 of the resolved skills dir. A `.acpx-skills-root` marker file (created `wx`) pins ownership so a hash collision fails loudly instead of silently flip-flopping links; torn/dead markers are reclaimed with bounded retries; a best-effort sweep removes stale roots (dangling links, dead owners, abandoned mid-creation roots past a 60s age gate).

## Test plan

- `pnpm run check` — 1,500 tests pass, coverage 93.71% lines
- New tests: capability gate rejection, synthetic-root layout on `session/new`, verbatim `--additional-dir`, lenient drop on `session/load` and lenient `createSession`, flag collection relative to `--cwd`, fs access to granted roots with outside paths still blocked

## Review round 2 (`23148a0`)

- **Normalize before client construction** — `normalizeSessionDirOptions` resolves relative `skillsDirs`/`additionalDirs` against the session cwd once, so the initial session, persistence, and reconnects agree (previously the first session resolved against host cwd).
- **Synthetic-root fs access** — `FileSystemHandlers.setAdditionalRoots` now takes a `skillTargets` map; reads through `<root>/.claude/skills/...` and `<root>/.agents/skills/...` are rewritten onto the real skills dir so `follow-within-root` containment holds.
- **Queued dir changes rejected** — a queued prompt carrying different `--skills-dir`/`--additional-dir` on a shared session client now fails with `SESSION_DIRS_IMMUTABLE` instead of silently keeping the creation-time roots.
- **Harness docs synced** — `agents/Codex.md` and `agents/Claude.md` document the flags per the harness-doc policy.
- **fs-callback boundary proof** — new test: reads through both synthetic links return real skill content, `..` escape to a sibling dir is rejected, a symlink inside the skills dir pointing outside is rejected, and writes through the link land on the real dir.

`pnpm run check` green: 0 failures, coverage 93.72%.


wtfsayo:
@clawsweeper re-review

steipete:
Thanks @wtfsayo for the implementation and runtime evidence. We are closing this combined proposal as not planned in its current form.

ACP additional workspace roots are a useful protocol feature, but `--skills-dir` adds a separate persistent cache, harness-specific directory layouts, symlink rewriting, and cleanup policy. It also exposes the skill source as a writable workspace root, which is broader than skill discovery. We do not want that additional filesystem convention in core.

The current patch also retains concrete correctness gaps despite green CI: previously granted roots survive some reconnect/capability-loss paths; filesystem grants are installed before failed path translations are filtered; omitted flags can reject ordinary reuse of a session with saved roots; and synthetic path normalization changes symlink-plus-parent traversal semantics. Relative roots also need to resolve consistently against the session cwd in embedded flows.

A focused generic `additionalDirectories` proposal can be considered separately, with one effective-root boundary, explicit read/write semantics, reconnect revocation, and production-path tests. The supplied skill-discovery trace is useful evidence, but does not establish those guarantees. Closing this PR rather than adopting the synthetic-root design.

PR #638, opened 2026-09-20
feat(runtime): expose session controls on shared sessions

saariuslystoned:
## What

`AcpRuntime` declares `setMode`, `setModel`, `setConfigOption` and `getCapabilities`,
and `createAcpRuntime()` implements them. `SharedAcpRuntime` did not, so an
application sharing a session with the CLI could submit turns, watch, cancel and
close, but had to shell out to `acpx` to change a mode, model or config option.

This implements all four on `SharedAcpRuntime`, with the same signatures and
return shapes as the in-process runtime. Everything below the runtime already
existed and is CLI-proven: the queue IPC messages, the owner handlers, and
`setSessionMode` / `setSessionModel` / `setSessionConfigOption`.

It also fixes two problems found while building it, both with regression tests.

## Design decisions

**Controls are owner-only; no direct-connection fallback.** `setSessionModel` and
friends fall back to `runSessionSet*Direct` when no queue owner is running. For a
shared client that is the wrong trade, so a control with no owner now fails with
`ACP_BACKEND_UNAVAILABLE` instead. The direct path passes no `resumePolicy`, so
`loadOrCreateRuntimeSession` may create a *fresh provider session* when the saved
one no longer loads — the same silent conversation swap that shared turns already
refuse via `resumePolicy: "same-session-only"`. It also contends for a session
another process owns. On a real agent that contention is not theoretical; against
an owner that still held the thread, codex-acp rejected the second connection
outright:

    QueueConnectionError: Internal error
      detailCode: 'QUEUE_CONTROL_REQUEST_FAILED'
      acp.data.details: 'thread 01a0bf39-…-e2752c4e8828 already has an active writer'

The CLI keeps its existing fallback; only the shared client fails closed.

**Authority is checked locally, and held all the way to the socket write.** The
interface documents "Authority gates dispatch", which is exactly what a shared
control can honour. `signal` and `assertActive` are checked on the way in, and
again immediately before the request is written to the owner's socket — the
control still has to read the owner record, resolve the advertised config key and
connect, and authority withdrawn across any of those awaits must still mean
nothing was sent. It is never consulted after the write, so a control the owner
accepted still settles its response and saved state. It does **not** gate controls the CLI or another application sends to
the same owner, and it cannot recall a control the owner already accepted.
`docs/shared-sessions.md` says so plainly. `startTurn` still rejects a per-turn
`assertActive`, unchanged.

**`getCapabilities` is derived, not invented.** No handle reports the controls
acpx implements; with a handle it adds `configOptionKeys` from the session
record's existing `acpx.config_options` — the same derivation `AcpxRuntime`
already used, now shared rather than duplicated. No new persisted field and no new
convention. `setConfigOption` resolves `key` through the same
`resolveSupportedConfigOptionId` as the in-process runtime, so `key` means the
same thing on both.

**Return shapes match the interface exactly.** `setMode`/`setModel` →
`Promise<void>`; `setConfigOption` → the adapter's `SetSessionConfigOptionResponse`.
No new fields.

## Fixes included

**`session/set_mode` was lost when a turn was active.** The queue owner's live
session controller updated `record.acpx` for model and config-option changes but
not for mode, so the turn's next checkpoint overwrote the requester's own record
write and the preference was silently dropped before the next reconnect replayed
it. `createAcpRuntime()` does not have this problem — its live controller sets
`desired_mode_id` on the turn state. This affects `acpx <agent> set-mode <id>`
today, not just the new API. Reproduced from the CLI alone with the mock agent, a
`set-mode` issued while a prompt streams:

    # before
    set mode -> mode set: plan
    desired_mode_id after turn: undefined
    # after
    set mode -> mode set: plan
    desired_mode_id after turn: "plan"

**Authority could be withdrawn while a control prepared its IPC request.** The
guard ran once on entry, but a control still awaits the config-key lookup, the
owner-record read (`readQueueOwnerRecord`) and the socket connect
(`connectToQueueOwner`) before `socket.write`. A `signal` aborted or an
`assertActive` invalidated across any of those still reached the owner and mutated
adapter state — contradicting the guarantee above. A dispatch-only guard is now
carried through the owner control helpers and invoked immediately before the
write. Thanks @clawsweeper for catching the deeper windows on this PR.

**Overlapping controls from one client could save a stale selection.** Each
control persists by reloading and rewriting the record after the owner replies, so
two concurrent calls could complete their writes in the opposite order and leave
the adapter on one value while the record kept the other. A shared client now runs
its controls one at a time per session. Controls from *separate* processes still
interleave the way separate `acpx` invocations do — that is pre-existing and would
need the owner to persist accepted state before replying, which is out of scope
here; it is documented in `docs/shared-sessions.md`.

## Docs

`docs/shared-sessions.md` no longer lists per-turn authority under what belongs to
`createAcpRuntime()`, and gains a "Session controls" section covering the
owner-only rule, the authority guarantee and its limits, and the serialization
limit.

## Commands run

    $ pnpm run check          # after the feature commit
    EXIT=0
    # tests 1410   # pass 1408   # fail 0   # skipped 2
    # tests 171    # pass 171    # fail 0          (coverage subset)

    $ pnpm run check          # after the serialization commit
    EXIT=0
    # tests 1411   # pass 1409   # fail 0   # skipped 2
    # tests 171    # pass 171    # fail 0          (coverage subset)

    $ pnpm run check:docs
    Checking formatting... All matched files use the correct format.  (43 files)
    markdownlint-cli2 ... Summary: 0 issues in 0 files
    built docs site: dist/docs-site

    $ .agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main
    # round 1: 1 x P2, concurrent control persistence  -> addressed
    # round 2: 1 x P2, authority window in setConfigOption -> addressed
    #          1 x P2, cross-process persistence -> see "Known limits"
    # round 3: 1 x P1, unserialized control write vs turn checkpoint
    #          -> pre-existing, see "Known limits"
    # after the ClawSweeper authority fix: pnpm run check exit 0,
    #   1411/1413 pass, 0 fail

`pnpm run mutate` was not run: its target is `src/cli/flags.ts`, untouched here.

## Tests

New cases in `test/shared-runtime.test.ts`, all against the real queue owner and
the built CLI:

- controls dispatched over a running owner's **live** connection during an active
  turn — asserts the adapter pid and the owner lease are unchanged, so no second
  connection was started
- an idle owner still serves a control and keeps its own provider session
- fail-closed with no owner: all three controls reject with
  `ACP_BACKEND_UNAVAILABLE`, and neither the record nor the adapter process moves
- authority rejection before send, for `assertActive` and an aborted `signal`, on
  all three controls, with the status unchanged afterwards
- authority withdrawn *during* a control's own config-key lookup
- authority withdrawn during owner lookup and socket connect, on all three
  controls — fails with `Missing expected rejection` without the dispatch guard
- `setConfigOption` response shape — exactly `["configOptions"]`, carrying the
  accepted value — plus `ACP_BACKEND_UNSUPPORTED_CONTROL` for an unadvertised key
- `getCapabilities()` and `getCapabilities({ handle })`, including an unknown
  record falling back to the base set
- `acpx set-mode` during an active turn survives the checkpoint (regression for
  the mode fix; fails against a build with the fix reverted)
- concurrent `setModel` calls persist the last accepted selection (regression for
  the serialization fix; fails 3/3 runs without it)

## Live verification

Real agent (`codex-acp`), three separate OS processes, session started from a
terminal so a human could attach to it:

    # terminal
    $ acpx --cwd $LIVE codex sessions new --name cli-owned
    $ acpx --cwd $LIVE codex prompt -s cli-owned 'Reply with exactly: READY'
    READY

    # a second process, using createSharedAcpRuntime()
    [B] getCapabilities({handle}) = {"controls":["session/set_mode","session/set_model",
        "session/set_config_option","session/status"],"configOptionKeys":["mode",
        "collaboration_mode","model","reasoning_effort","fast-mode"]}
    [B] model before = gpt-6-astra
    [B] withheld authority rejected: host withheld authority
    [B] model unchanged after rejection = gpt-6-astra
    [B] setModel(gpt-5.6-sol) ok model now = gpt-5.6-sol
    [B] setConfigOption(reasoning_effort=high) ok response keys ["configOptions"] reasoning_effort = high
    [B] setMode(read-only) ok

    # back in the terminal
    $ acpx --cwd $LIVE --format json codex sessions show cli-owned
    current_model_id: gpt-5.6-sol
    desired_mode_id : read-only
    desired_config_options: {"reasoning_effort": "high"}

    # third process, after the owner's idle TTL expired
    [C] setMode -> ACP_BACKEND_UNAVAILABLE: No running owner holds this shared session,
        so session/set_mode was not sent. ...
    [C] model still = gpt-5.6-sol

Authority revoked *during* IPC preparation, on the production path, against a real
`codex-acp` owner with a prompt turn streaming — the timing the earlier transcript
did not cover:

    [R] model before = gpt-6-astra
    [R] setModel: rejected after 2 checks -> host revoked authority mid-dispatch
    [R] setMode: rejected after 2 checks -> host revoked authority mid-dispatch
    [R] setConfigOption: rejected after 2 checks -> host revoked authority mid-dispatch
    [R] model after  = gpt-6-astra
    [R] unchanged = true

    # and from the terminal, nothing reached the adapter or the record
    $ acpx --cwd $LIVE --format json codex sessions show revoke-proof
    current_model_id: gpt-6-astra
    desired_mode_id : None
    desired_config_options: null

## Known limits, not addressed here

- **Control persistence is not serialized.** Every control saves accepted state by
  reloading the session record and writing it back from the requesting process.
  That write is not coordinated with another process's control, nor with the
  owner's own checkpoints, so it can land on a stale snapshot. This is
  byte-for-byte what `main` already does for every `acpx set` / `set-mode`
  (`git show main:src/session/execution/session-control.ts`), and
  `writeSessionRecord` has no lock or CAS. `acquireSessionTurn` cannot serve here:
  the owner holds it for a whole turn, and controls are answered during one. A
  correct fix is owner-side persistence before the IPC reply, or an atomic record
  update — an architecture change I did not want to fold into a feature PR. It is
  called out in the helper's doc comment and in `docs/shared-sessions.md`, and filed
  separately as [#639](https://github.com/openclaw/acpx/issues/639) with the two
  directions I can see. Happy to implement whichever you prefer.

  Worth noting this PR removes one instance of that class rather than adding one:
  before it, the owner's checkpoint reliably erased a mode set during a turn.
- Between turns the owner serves a control on a connection of its own, through
  `runSessionSet*Direct` with no `resumePolicy`. Unchanged by this PR, but it is
  the same create-a-different-session hazard the shared client now refuses, and
  with `codex-acp` it fails outright. Filed as
  [#640](https://github.com/openclaw/acpx/issues/640) with a CLI-only reproduction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


saariuslystoned:
Thanks — the authority finding is correct, and it was worse than the report says: the control path never plumbed `signal` down to `runQueueOwnerRequest` either, so the existing pre-write `signal` check there was dead for controls. Fixed in `522aa52`.

**What changed.** A dispatch-only guard is now carried through the owner control helpers — `setSession*OnOwner` → `trySet*OnRunningOwner` → `submitSet*ToQueueOwner` → `submitControlToQueueOwner` → `runQueueOwnerRequest` — and invoked immediately before `socket.write`, next to the existing `signal` check. It is never consulted afterwards, so a dispatched control still settles its response and saved state, as `AcpRuntime` documents. The parameter is optional, so the CLI callers are unchanged.

**Regression test.** `authority withdrawn during owner lookup and connect blocks the write` covers all three controls with an `assertActive` that passes on entry and throws on the recheck, then asserts the status is byte-identical afterwards. Removing just the guard call at the write point and rebuilding:

```text
not ok 1 - authority withdrawn during owner lookup and connect blocks the write
  error: 'Missing expected rejection.'
```

So the request really was being sent before; it is not now.

**Production-path proof**, which is what you asked for. Real `codex-acp` owner, a prompt turn streaming, authority revoked while the control is still locating the owner and connecting:

```text
[R] model before = gpt-6-astra
[R] setModel: rejected after 2 checks -> host revoked authority mid-dispatch
[R] setMode: rejected after 2 checks -> host revoked authority mid-dispatch
[R] setConfigOption: rejected after 2 checks -> host revoked authority mid-dispatch
[R] model after  = gpt-6-astra
[R] unchanged = true
```

and from a terminal, confirming nothing reached the adapter or the record:

```text
$ acpx --cwd "$DIR" --format json codex sessions show revoke-proof
current_model_id: gpt-6-astra
desired_mode_id : None
desired_config_options: null
```

`pnpm run check` is green on the new head: exit 0, 1411/1413 pass, 0 fail, 2 skipped. The PR body has the full command list and both transcripts.

**On the persistence race** — agreed it is pre-existing and should stay tracked separately; it is now [#639](https://github.com/openclaw/acpx/issues/639), with the file/line evidence and two proposed directions. While proving this fix I also reproduced a second, unrelated pre-existing bug from the CLI alone: `acpx codex set model` fails with `already has an active writer` when the queue owner is alive but idle, while working both with no owner and during a live turn. That is [#640](https://github.com/openclaw/acpx/issues/640). Happy to take either one once a maintainer picks a direction.

saariuslystoned:
Fixed in `9e32522`. The finding is right, and it contradicted evidence I had already gathered myself — while investigating [#640](https://github.com/openclaw/acpx/issues/640) I watched `acpx codex set model` succeed with no owner running, which is exactly the direct path that does not leave an owner behind. The diagnostic should have said so.

Before rewording I checked the claim rather than assuming it:

```text
$ acpx --cwd "$DIR" codex sessions new --name noowner-claim
01a0bf8b-023e-74a0-bbfb-60b1f392afd9
# owner before: no owner
$ acpx --cwd "$DIR" codex set model gpt-5.6-luna -s noowner-claim
model set: gpt-5.6-luna
# owner after: no owner left behind
```

So the CLI command applies the selection over a temporary connection and leaves no owner, and a shared caller who follows the old advice and retries would get `ACP_BACKEND_UNAVAILABLE` again.

The error is now:

> No running owner holds this shared session, so `session/set_mode` was not sent. Submit a turn to start an owner, or apply the change with the equivalent acpx command, which connects directly instead of starting one.

and `docs/shared-sessions.md` now says submitting a turn starts an owner, while the `acpx` command is an alternative that applies the control over a temporary direct connection without leaving an owner behind for the next shared control.

`pnpm run check` exit 0 (1411/1413 pass, 0 fail, 2 skipped) and `pnpm run check:docs` exit 0 on the new head.

On the one open note in the review — post-dispatch settlement timing not being independently demonstrated in the transcript: agreed that is structural rather than observed. The guard is invoked immediately before `socket.write` and never referenced afterwards, and the caller's `signal` is deliberately not attached as a post-dispatch listener, so there is no path by which a dispatched control can be revoked. If you would like that shown empirically rather than argued, say the word and I will add a test that revokes authority after the write and asserts the response and saved state still settle.

saariuslystoned:
@clawsweeper re-review

steipete:
Landed as 22032d219e28cd7cc699cf2504dade007ba3de3e from reviewed head beae479e0a4ea84be3815527d27f4fcca3a810e8. Thanks @saariuslystoned!

Shared clients now use the existing owner's mode/model/config controls, with final-dispatch authority checks and accepted-state settlement after dispatch. The landing adds stronger boundary tests and clarifies saved config-key discovery, queue error behavior, and mixed-version persistence.

Verification:

- `pnpm run check`: 1,493 passed, two skipped, zero failures; all 171 coverage tests and the declared coverage thresholds passed.
- `pnpm run check:docs`: passed.
- `.agents/skills/autoreview/scripts/autoreview --mode local --base origin/main`: clean, no accepted/actionable findings. The final rebase preserved the complete validated file tree. The fresh ClawSweeper revision-7 review also reported no findings.
- Independent source-blind checks through the built runtime and CLI with a synthetic ACP peer: 50 assertions passed across active and idle controls, capabilities, no-owner rejection, model validation, ordering, authority revocation, same-session identity, and cleanup.
- Final-head CI: all 12 jobs green, including Node 22/24/26 tests, mutation checks, and Slophammer: https://github.com/openclaw/acpx/actions/runs/35529246891.

The first local full run hit the pre-existing three-second watch-rotation watchdog. The complete run passed after rebasing onto the reviewed test repair in https://github.com/openclaw/acpx/pull/643. These maintainer behavior probes are synthetic local proof; no new real-provider claim is made.

The legacy caller-write race remains best-effort compatibility, with the updated-caller and updated-owner requirement documented explicitly.

PR #641, opened 2026-09-20
fix: serialize abandoned session turn recovery

steipete:
Two processes reclaiming an abandoned session-turn marker can both become owners: the old read/compare/unlink sequence lets a delayed reclaimer delete the newly admitted writer's marker. The same gap exists between marker removal and completion of release. This can overlap persistent session checkpoints and duplicate concurrent `sessions ensure` creation.

Use the already installed fs-safe lock primitive as a private guard held from marker admission through final cleanup. Keep the existing marker encoding and exclusive-create/hardlink fallback. Only a definitely dead PID authorizes guard recovery; age, permission errors and malformed guard contents do not. Waiting remains cancellable without detaching a future acquisition, and cancellation after admission does not release a writer before its checkpoint. Disposal is serialized, idempotent and retryable after cleanup failures. Failed admission settles cleanup and guard release while preserving all errors.

The test changes also isolate USERPROFILE with HOME on Windows and move the close-checkpoint admission barrier to the new acquisition boundary. Existing legacy live markers remain respected. Old binaries do not honor the additional guard, so the stronger exclusion applies to updated cooperating participants. Ambiguous guards or interrupted fs-safe reclamation residue remain fail-closed and require externally exclusive cleanup; this patch does not age-delete them or change queue lease/socket ownership.

Validation: three initial regressions fail on the base, including actual separate processes. Review also exposed a legacy acquirer winning the post-recovery publication race; its new regression failed before the final repair and now passes with ordinary contention handling. The 32 targeted ownership and close-checkpoint tests pass in the final full gate, including a real owner killed with both lock files present, aged partial no-hardlink publication, cancelled admission, release gaps, retryable cleanup, and final close checkpoints. The final `pnpm run check` and `pnpm run check:docs` logs show 1,423 passed / 2 skipped plus the declared coverage and docs gates on AWS Crabbox `cbx_77f655cf7d4d`, run `run_8959d365fde059dfff6a2acca3b7eecd`. The local wrapper was interrupted after the completed docs output, so its terminal broker receipt is unavailable. Exact-head GitHub CI will provide the final merge gate. Direct changed Markdown checks pass. Final isolated review is scoped-clean through P2 at candidate `85274170549b912d9906d8d260926b1ab8c82144`; commit hooks left every reviewed file unchanged.

<details>
<summary>Built CLI before/after proof</summary>

On macOS, two real built CLI processes invoke the same command in an isolated home:

```sh
node dist/cli.js --cwd "$fixture" --agent "$peer_command" --format json --deny-all sessions ensure --name guard-proof
```

A synthetic native ACP peer records actual protocol requests and holds the first `session/new` response behind a barrier. A local preload hook pauses only the first process's stale-marker unlink, then records when the second process enters turn acquisition. It does not replace the lock implementation, control RPCs, or session persistence.

The pre-fix build fails: while the first process is still paused at stale unlink, the second emits two native requests (`initialize` and `session/new`). With the repair, the second process remains blocked through stale removal and through the first process's held creation response. Releasing that response yields:

```json
{"pass":true,"blockedDuringStaleRemoval":true,"blockedDuringCreation":true,"nativeCreates":1,"created":[true,false],"sameRecord":true,"exits":[0,0],"peerExited":true,"markersRemoved":true}
```

The driver verifies both returned records, the actual native request count, process exits, and removal of both marker and guard. Temporary state and peers are cleaned up. Candidate built CLI SHA-256: `d382ece296d3573e02e8a206ec66c174a2b9c59497d43de020bccd66cd2bbde2`.

</details>


steipete:
Maintainer disposition: accept the documented recovery boundary for this serialization repair. The stronger exclusion applies to updated cooperating processes; existing live legacy markers are still respected, including a legacy process winning marker creation during recovery. This is not a claim that old reclaimers gain the new protocol's protection.

Ambiguous guard contents and interrupted reclamation residue remain preserved. Cleanup requires excluding competing processes; age is not authority to delete a possible live owner's guard. That behavior is explicit in the session documentation and covered by cancellation/retention regressions. A different automatic-recovery protocol would be separate work with its own cross-platform proof, rather than an ungated fallback in this patch.

The review reports no introduced correctness findings and accepts the built-CLI proof. Its availability and upgrade caveats remain recorded; this resolves the requested maintainer decision for the scoped fix. Landing remains contingent on the exact-head CI gate.

steipete:
Landed as 970a804f030b6c2ab19fbadb13db32db01ba3a38 from reviewed head 85274170549b912d9906d8d260926b1ab8c82144. All 12 CI jobs passed in run 35520547009, including the Node 22/24/26 suites, mutation and Slophammer gates; CodeQL also passed. Main is synchronized and clean.

The existing filesystem lock primitive now holds exclusion across abandoned-marker recovery, admitted work and cleanup. Four before-failing race/compatibility regressions, 32 targeted ownership/close cases, and real built-CLI before/after proof cover the repair. Final full logs show 1,423 passed / 2 skipped plus declared coverage/docs; exact-head CI independently confirms the gates despite the interrupted local broker receipt. The final isolated review is scoped-clean through P2.

ClawSweeper accepted the production-path proof and reported no introduced correctness finding. Its documented availability and mixed-version caveats are retained, with the maintainer disposition recorded above: ambiguous guard residue requires externally exclusive cleanup, and stronger exclusion applies to updated cooperating implementations. Queue lease reclamation and control-state ownership remain separate audit work.

